PR #55, opened 2025-02-14
Initial code consolidation work

sonals:
Initial work to consolidate regex usage in various files

1. Consolidate two utils.h into a common utils.h
2. Initial work to consolidate regex usage in various files

NA

Remove duplicate files and start putting down infra for creating repository of regex.

None

Build and ctest validated

NA


src/cpp/aiebu/src/common/utils.h
```diff
@@ -164,12 +164,26 @@ vector_of_string_to_vector_of_char(const std::vector<std::string>& args)
...
+  R_BRACK_RE,
 };
 
 std::regex get_regex(const std::vector<fragment>& pattern);
 
+constexpr unsigned hexbase = 0x10;
+
+template <typename INTT> INTT to_int(const std::string &token) {
+  const unsigned long long result = (std::is_same<INTT, uint64_t>::value) ? std::stoull(token, nullptr, hexbase) :
+    std::stoul(token, nullptr, hexbase);
```
stsoe: This doesn't look very flexible.  It converts with base16, if that is really the intend, the function name should reflect this.  Else, I think the base should set to `0` to autodetect hexadecimal (0x), octal (0), or otherwise decimal.

The formatting of the template and function names makes it very hard to read. Better
```
template <typename IntType>
IntType 
to_int(const std::string& token)
{
  ...
}
```

src/cpp/aiebu/src/common/utils.h
```diff
@@ -164,12 +164,26 @@ vector_of_string_to_vector_of_char(const std::vector<std::string>& args)
...
+  INTT max = 0;
+  max = ~max;
```
stsoe:
```suggestion
  auto max = std::numeric_limits<IntType>::max();
```

src/cpp/aiebu/src/common/utils.h
```diff
@@ -164,12 +164,26 @@ vector_of_string_to_vector_of_char(const std::vector<std::string>& args)
...
+  if (result > max)
+    throw error(error::error_code::invalid_asm, "Value " + token + " is out of range");
+  return static_cast<INTT>(result);
```
stsoe:
```suggestion
  if (result > max)
    throw error(error::error_code::invalid_asm, "Value " + token + " is out of range");
    
  return static_cast<INTT>(result);
```

stsoe (changes requested):

src/cpp/aiebu/src/common/utils.h
sonals: Done

src/cpp/aiebu/src/common/utils.h
sonals: Done

src/cpp/aiebu/src/common/utils.h
sonals: Done

src/cpp/aiebu/src/common/utils.h
```diff
@@ -144,11 +145,47 @@ inline std::string get_pagelabel(const std::string& label)
...
+}
+}
+
+enum class fragment{
+  BEGIN_ANCHOR_RE,
+  END_ANCHOR_RE,
+  HEX_RE,
+  L_BRACK_RE,
+  R_BRACK_RE,
+};
```
stsoe: enum class enumerators should be lower case
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#enum5-dont-use-all_caps-for-enumerators

src/cpp/aiebu/src/common/utils.h
```diff
@@ -144,11 +145,47 @@ inline std::string get_pagelabel(const std::string& label)
...
+
+std::regex get_regex(const std::vector<fragment>& pattern);
+
+constexpr unsigned hexbase = 0x10;
+
+template <typename IntType>
+IntType
+to_uinteger(const std::string& token) {
+  const unsigned long long result = (std::is_same<IntType, uint64_t>::value) ? std::stoull(token, nullptr, 0) :
```
HimanshuChoudhary-Xilinx: should we put stoul and stoull in try catch ? as they throw invalid_argument or out_of_range exception

stsoe: > should we put stoul and stoull in try catch ? as they throw invalid_argument or out_of_range exception

Only catch if you want to handle the error.  If stoul throws, it is better to propagate the error up to main().

PR #65, opened 2025-02-26
Pre-generate header/doc files based on isa-spec.yaml

stsoe:
When isa-spec.yaml is updated, pre-generated files must be recreated through isa-spec-gen target.

```
cmake --build . --config Release --target isa-spec-gen
```

This PR moves the generation from src/cpp/aiebu/src/CMakeLists.txt into
- specification/CMakeLists.txt and
- specification/{aie2,aie2ps}/CMakeLists.txt

The pre-generated files are created in the source tree where they are checked in.

The isa-spec-gen target is not invoked as part of a regular build, but can be run by specifying the target.  If the files are different they will have to be added and comitted along with the new isa-spec.yaml file.

Still to-do is to have a default target that compare the timestamp of isa-spec.yaml against that of the pre-generated files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,9 +9,10 @@ add_subdirectory(src/cpp/aiebu)
 
 if (AIEBU_FULL STREQUAL "ON")
   add_subdirectory(src/python)
-  add_subdirectory(specification)
   add_subdirectory(scripts)
 endif()
+
+add_subdirectory(specification)
 
 add_subdirectory(lib)
 add_subdirectory(test)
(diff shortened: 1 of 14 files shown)


sonals (approved):
"The pre-generated files are created in the source tree where they are checked in." 
If I regenerate the files, will they be produced in build directory or in the source directory?

stsoe:
> "The pre-generated files are created in the source tree where they are checked in." If I regenerate the files, will they be produced in build directory or in the source directory?

In the source tree and changes (if any) will show up in `git status`.

PR #68, opened 2025-02-27
Store 'raw' md5sum value in note.xrt.uid

sonals:
Currently we are generating md5sum, converting it to ASCII hex and then storing the hex string as raw value in note.xrt.uid. Hence the value stored is binary representation of the hex string not the original md5sum. 

The issue has been there from the beginning

Fixed by md5sum generation and string representation in aiebu::uid_md5 class.

Starting with this change, the note.xrt.uid will have a different value for the same txn and ctrlpkt binary. 

Compile and ctest validated

None


sonals:
Will regenerate the PR

PR #69, opened 2025-02-27
Store 'raw' md5sum value in note.xrt.uid

sonals:
Currently we are generating md5sum, converting it to ASCII hex and then storing the hex string as raw value in note.xrt.uid. Hence the value stored is binary representation of the hex string is not the original md5sum which is wrong.

The issue has been there from the beginning

Fixed by updating the md5sum generation and string representation in aiebu::uid_md5 class.

None

Build and ctest validated

NA


src/cpp/aiebu/src/elf/elfwriter.cpp
```diff
@@ -183,38 +183,38 @@ add_text_data_section(std::vector<writer>& mwriter, std::vector<symbol>& syms)
 {
   for(auto buffer : mwriter)
```
stsoe, line 184: This is a lot of copying,  should it be `auto&`

sonals: We have the copy inefficiency in several places. Will consider cleaning this up in a separate PR.

stsoe (approved):

PR #84, opened 2025-04-23
aiebu changes for cross compilation for embedded support

HimanshuChoudhary-Xilinx:
aiebu changes for cross compilation for embedded support
1. we have added pre-generated stx_save_restore_map_prebuilt.h only for cross compile
2. TODO: to add infra to generate stx_save_restore_map.h on fly during build


cmake/settings.cmake
```diff
@@ -11,6 +11,14 @@ set(AIEBU_VERSION_RELEASE 202510)
 SET(AIEBU_VERSION_MAJOR 1)
 SET(AIEBU_VERSION_MINOR 0)
 
+message("-- Host system processor is ${CMAKE_HOST_SYSTEM_PROCESSOR}")
+message("-- Target system processor is ${CMAKE_SYSTEM_PROCESSOR}")
+
+set(AIEBU_NATIVE_BUILD "yes")
+if (NOT ${CMAKE_SYSTEM_PROCESSOR} STREQUAL ${CMAKE_HOST_SYSTEM_PROCESSOR})
+  set(AIEBU_NATIVE_BUILD "no")
+endif()
```
stsoe: I would recommend using `set(AIEBU_NATIVE_BUILD TRUE)`  or `set(AIEBU_NATIVE_BUILD 1)` and then simply use `if (AIEBU_NATIVE_BUILD)` as opposed to `if (${VAR} STREQUAL ...)`

src/cpp/preprocessor/aie2/stx_save_restore_map_prebuilt.h
```diff
@@ -0,0 +1,18 @@
...
+// Copyright (C) 2024, Advanced Micro Devices, Inc. All rights reserved.
+
+#ifndef AIEBU_STX_PREEMPTION_FILES_H
+#define AIEBU_STX_PREEMPTION_FILES_H
+
+#include <map>
+#include <vector>
+#include <cstdint>
+
+const std::map<uint32_t, std::pair<std::vector<uint8_t>, std::vector<uint8_t>>> stx_save_restore_map = {
```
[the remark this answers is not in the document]
stsoe: I think we should address this warning.  I assume `stx_save_restore_map`  is used at runtime?
Move the constant as a static variable into a function where you access it a runtime.   If construction throws, then application will catch it.

lib/CMakeLists.txt
```diff
@@ -9,4 +9,6 @@ if (${AIEBU_AIE_RT_BIN_DIR} STREQUAL ${AIEBU_BINARY_DIR})
   aiebu_add_subdirectory_disable_install_target(aie-rt/driver)
 endif()
 
-add_subdirectory(src)
+if (${AIEBU_NATIVE_BUILD})
```
stsoe:
```suggestion
if (AIEBU_NATIVE_BUILD)
```

If you write if(${FOO}), CMake first expands ${FOO} before evaluating the condition. This can lead to subtle bugs, especially if FOO expands to a value that is itself the name of another variable or an empty string

src/cpp/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 # SPDX-License-Identifier: MIT
 # Copyright (C) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
-add_subdirectory(utils)
+
+# In arm compiler we dont have static libs so not building utility
+if (${AIEBU_NATIVE_BUILD})
```
stsoe:
```suggestion
if (AIEBU_NATIVE_BUILD)
```

src/cpp/CMakeLists.txt
```diff
@@ -48,13 +52,20 @@ target_compile_definitions(aiebu_library_objects
   -DAIEBU_EXPORTS
   )
 
+if (${AIEBU_NATIVE_BUILD})
```
stsoe:
```suggestion
if (AIEBU_NATIVE_BUILD)
```

test/CMakeLists.txt
```diff
@@ -6,7 +6,10 @@ project(tests CXX C)
 if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
   add_subdirectory(aie2ps-ctrlcode)
 endif()
-add_subdirectory(aie2-ctrlcode)
+
+if (${AIEBU_NATIVE_BUILD})
```
stsoe:
```suggestion
if (AIEBU_NATIVE_BUILD)
```

stsoe (changes requested):

stsoe (approved):
It's annoying that clang-tidy spews tens for warnings about the constants.   You can prevent this by enclosing the code within 
```
// NOLINTBEGIN 
... 
// NOLINTEND
```
But whatever, it is fine as is.

chvamshi-xilinx (approved):

PR #86, opened 2025-05-02
aie4 changes

HimanshuChoudhary-Xilinx:
Aiebu change to support aie4 target
1. added new target asm_aie4
2. aie2ps and aie4 have 2 things different
     a. patch57 schema and control packet schema
     b. actor_id's for tiles/mm2s/s2mm
3. created assembler_state base class and aie2ps and aie4 derived class. We have create_assembler_state virtual function which will return aie4 or aie2ps assembler_state shared ptr.
4. created aie4 encoder class from aie2ps encoder as only patch57 and assember_state is different.

NA

NA

#### USAGE
```
> aiebu-asm -t aie2ps -c ..\test\cpp_test\aie2ps\eff_net_coal\ml_asm\merged_control.asm -L ..\test\cpp_test\aie2ps\eff_net_coal\ml_asm -L ..\test\cpp_test\aie2ps\eff_net_coal\asm\ -o controlcode.elf

```

### DIFF
--- a/src/cpp/common/assembler_state.h
+++ b/src/cpp/common/assembler_state.h
@@ -130,12 +130,11 @@ protected:
 
   //std::unordered_map<std::string, ActionId> actor_id;
 
-  virtual std::unordered_map<std::string, ActionId>& get_actor_id_map() const = 0;
-  uint32_t get_actor(const std::string& prefix, const std::string& s) const
+  virtual std::unordered_map<std::string, uint32_t>& get_actor_id_map() const = 0;
+  uint32_t get_actor(const std::string& s) const
   {
-    std::unordered_map<std::string, ActionId>& actor_id = get_actor_id_map();
-    uint32_t actor = std::stoi(s.substr(actor_id.at(prefix).actor_start));
-    return actor_id.at(prefix).base_actor_offset + actor;
+    std::unordered_map<std::string, uint32_t>& actor_id = get_actor_id_map();
+    return actor_id.at(s);
   }
 
   assembler_state(std::shared_ptr<std::map<std::string, std::shared_ptr<isa_op>>> isa,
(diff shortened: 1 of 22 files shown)


stsoe (commented):
There are other fundamental changes, such as pointer semantics.  Please describe the changes in the PR  summary, rather than just a one liner.

src/cpp/common/assembler_state.h
```diff
@@ -200,7 +198,77 @@ class assembler_state
...
+  symbol::patch_schema get_control_packet_patching() const override { return symbol::patch_schema::control_packet_57; }
 };
 
+class assembler_state_aie4 : public assembler_state
+{
+  std::unordered_map<std::string, ActionId>&
+  get_actor_id_map() const override
+  {
+    static std::unordered_map<std::string, ActionId> actor_id = {
+      {"mm2s", {5, 6}},               //NOLINT
```
xuhz: aie4, if tile type is not specified, it is an error.
so line 246-247 should be removed

src/cpp/common/assembler_state.h
```diff
@@ -200,7 +198,77 @@ class assembler_state
...
+      {"s2mm", {5, 0}},               //NOLINT
+      {"tile_mm2s", {10, 6}},         //NOLINT
+      {"tile_s2mm", {10, 0}},         //NOLINT
+      {"shim_mm2s", {10, 6}},         //NOLINT
```
xuhz: please see table 3-23 in latest spec
this part now is very ugly ...
memtile, the mm2s actor id has a hole
shimtile, actor id of s2mm 0 & 1 are not contiguous
there is also a trace_s2mm channel

check to see whether the data structure define can handle all these case.

xuhz (changes requested):

stsoe (approved):
LGTM

chvamshi-xilinx (approved):

src/cpp/common/assembler_state.h
HimanshuChoudhary-Xilinx: checked and fixed

src/cpp/common/assembler_state.h
HimanshuChoudhary-Xilinx: done

xuhz (approved):

PR #90, opened 2025-05-14
Bd

sonals:
#### Problem solved by the commit
Add support for dumping aie2p control packets. This feature is tied to -D switch of aiebu-dump.

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
NA

#### How problem was solved, alternative solutions (if any) and why they were rejected
Added new code to parse packet header and control header and write them out in a human readable format.

#### Risks (if any) associated the changes in the commit
None

#### What has been tested and how, request additional testing if necessary
Compile and ctest validated

#### Documentation impact (if any)
NA

PR #95, opened 2025-06-05
Add support for AIE2 & AIE2P ctrlpkt dump from ctrlpkt binary file

donwalkarsoham:
AIE2 and AIE2P control packet binaries are packaged differently.
1. For AIE2 - Each control packet entry in the control packet binary is 8 words aligned. That is, every control packet starts after 8 words from the start address of the previous control packet.
Example - control_packet_2 = Address of control_packet_1 + 8 words.

2. For AIE2P - Control packets are contiguous. That is each control packet entry in the control packet binary starts right after the previous control packet entry ends. 
Example - Start address of control_packet_2 = Start address of control_packet_1 + size of control_packet_1

Feature - Added support to display control packets for AIE2 and AIE2P using above specs when a control packet binary is provided.

Bug Fix - The existing control packet dump from ELF file doesn't follow above architecture specific rule but treats all control packets as per rule 2 above. This has been fixed.

https://github.com/Xilinx/aiebu/pull/90

1> Added additional buffer type "aiebu_assembler::buffer_type::blob_control_packet_aie2" for AIE2 specific control packet as this is a distinct category of control packet specific to AIE architecture without changing the existing one ("aiebu_assembler::buffer_type::blob_control_packet" ) to maintain backward compatibility.
But I think to further enhance nomenclature and clarity we can add "aiebu_assembler::buffer_type::blob_control_packet_aie2p" to have same functionality as existing "aiebu_assembler::buffer_type::blob_control_packet" and eventually deprecate "aiebu_assembler::buffer_type::blob_control_packet".
2> Added changes to detect control packet type from given ELF or from given control packet binary and accordingly display the control packet
3> "-d" option of aiebu-dump is extended to support both ELF input and control packet binary input.

N/A

1> All the build time tests have passed.
100% tests passed, 0 tests failed out of 88

2> Tested AIE2 and AIE2P control packet binaries and compared them with the transparser tool output which has the arch specific dumping enabled.

2.1 - Input is AIE2 Control packet Binary-
 **./aiebu-dump <path_to_ctrl_pkt0.bin>/ctrl_pkt0.bin -d**

aie2-ctrlpkt
PH {#4, R[0?], C[0?]}
CH {@0x1d000, [4], P}
   [0] 0x800
   [1] 0x0
   [2] 0x0
   [3] 0x0
PH {#4, R[0?], C[0?]}
CH {@0x1d010, [4], P}
   [0] 0x80000000
   [1] 0x2000000
   [2] 0x0
   [3] 0x2001001
PH {#2, R[0?], C[0?]}
CH {@0xa0000, [4], P}
   [0] 0x800
   [1] 0x20000
   [2] 0x0
   [3] 0x0
PH {#2, R[0?], C[0?]}
...

2.2 - Input is AIE2P Control packet Binary-
**./aiebu-dump <path_to_ctrl_pkt0.bin>/ctrl_pkt0.bin -d**

aie2p-ctrlpkt
PH {#8, R[0?], C[0?]}
CH {@0x32000, [1], P}
   [0] 0x0
PH {#8, R[0?], C[0?]}
CH {@0xef00, [4], P}
   [0] 0x0
   [1] 0x0
   [2] 0x0
   [3] 0x0
PH {#8, R[0?], C[0?]}
CH {@0xef10, [4], P}
   [0] 0x0
   [1] 0x0
   [2] 0x0
   [3] 0x0
PH {#8, R[0?], C[0?]}
CH {@0xef20, [4], P}
   [0] 0x0
   [1] 0x0
   [2] 0x0
   [3] 0x0
...

3> Input is ELF for AIE2
**./aiebu-dump <output.elf>/output.elf -D**

aie2-elf

Section[2]: .ctrltext   Size: 1232B

    .attach_to_group 0
    .attach_to_group 1
    .attach_to_group 2
    .attach_to_group 3
    .attach_to_group 4

XAIE_IO_BLOCKWRITE              @0x201d000, [8]
XAIE_IO_BLOCKWRITE.0            0x6
XAIE_IO_BLOCKWRITE.1            0x0
XAIE_IO_BLOCKWRITE.2            0x0
XAIE_IO_BLOCKWRITE.3            0x0
XAIE_IO_BLOCKWRITE.4            0x80000000
XAIE_IO_BLOCKWRITE.5            0x2000000
XAIE_IO_BLOCKWRITE.6            0x300007
XAIE_IO_BLOCKWRITE.7            0x2000000
XAIE_IO_CUSTOM_OP_DDR_PATCH     @0x201d004, #3, +0x0
XAIE_IO_WRITE                   @0x201d214, 0x30000
XAIE_IO_BLOCKWRITE              @0x201d020, [8]
XAIE_IO_BLOCKWRITE.0            0x3
XAIE_IO_BLOCKWRITE.1            0x0
...

Section[3]: .ctrldata   Size: 1248B

PH {#4, R[0?], C[0?]}
CH {@0x1d000, [4], P}
   [0] 0x800
   [1] 0x0
   [2] 0x0
   [3] 0x0
PH {#4, R[0?], C[0?]}
CH {@0x1d010, [4], P}
   [0] 0x80000000
   [1] 0x2000000
   [2] 0x0
   [3] 0x2001001
PH {#2, R[0?], C[0?]}
CH {@0xa0000, [4], P}
   [0] 0x800
   [1] 0x20000
   [2] 0x0
   [3] 0x0
PH {#2, R[0?], C[0?]}
CH {@0xa0010, [4], P}
   [0] 0x0
   [1] 0x0
   [2] 0x0
   [3] 0x81418040
PH {#4, R[0?], C[0?]}
CH {@0x14000, [1], P}
   [0] 0x0
...

### DIFF
--- a/src/cpp/analyzer/packets.cpp
+++ b/src/cpp/analyzer/packets.cpp
@@ -9,6 +9,11 @@
 
 namespace aiebu {
 
+// Size of word in bytes
+constexpr unsigned int word_size = 4;
+// For AIE2 control packets are 8 words (8words * 4bytes/word = 32bytes) aligned
+constexpr unsigned int ctrlpkt_offset_aie2 = 8 * word_size;
+
 char control_packet_operations_map(uint32_t op)
 {
   static const std::array<char, 4> optable = {'P', 'R', 'W', 'U'};
--- a/src/cpp/analyzer/packets.cpp
+++ b/src/cpp/analyzer/packets.cpp
@@ -43,20 +43,20 @@ packets::get_dump() const
 {
   std::stringstream stream;
   size_t offset = 0;
-  while (offset < m_size) {
-    offset = serialize(stream, offset);
-  }
-  return stream.str();
-}
-
-std::string
-packets::get_dump_aie2() const
-{
-  std::stringstream stream;
-  size_t offset = 0;
-  while (offset < m_size) {
-    serialize(stream, offset);
-    offset += ctrlpkt_offset_aie2;
+  // check control packet type
+  if (m_buffer_type == aiebu::aiebu_assembler::buffer_type::blob_control_packet) {
+    // AIE2P control packet
+    while (offset < m_size) {
+      offset = serialize(stream, offset);
+    }
+  } else if (m_buffer_type == aiebu::aiebu_assembler::buffer_type::blob_control_packet_aie2) {
+    // AIE2 control packet
+    while (offset < m_size) {
+      serialize(stream, offset);
+      offset += ctrlpkt_offset_aie2;
+    }
+  } else {
+    stream << "Unknown buffer type for control packet dump.\n";
   }
   return stream.str();
 }
--- a/src/cpp/analyzer/packets.cpp
+++ b/src/cpp/analyzer/packets.cpp
@@ -49,4 +49,16 @@ packets::get_dump() const
   return stream.str();
 }
 
+std::string
+packets::get_dump_aie2() const
+{
+  std::stringstream stream;
+  size_t offset = 0;
+  while (offset < m_size) {
+    serialize(stream, offset);
+    offset += ctrlpkt_offset_aie2;
+  }
+  return stream.str();
+}
+
 }
--- a/src/cpp/analyzer/packets.h
+++ b/src/cpp/analyzer/packets.h
@@ -4,6 +4,7 @@
 #include <string>
 #include <cstdint>
 #include <iostream>
+#include "aiebu/aiebu_assembler.h"
 
 namespace aiebu {
 
@@ -11,14 +12,16 @@ class packets {
 private:
   const char *m_buffer;
   uint64_t m_size;
+  aiebu::aiebu_assembler::buffer_type m_buffer_type;
 
 private:
   size_t serialize(std::ostream &stream, size_t offset) const;
 
 public:
-  packets(const char *buffer, uint64_t size) : m_buffer(buffer), m_size(size) {}
+  packets(const char *buffer, uint64_t size, aiebu::aiebu_assembler::buffer_type buffer_type)
+    : m_buffer(buffer), m_size(size), m_buffer_type(buffer_type) {}
+
   std::string get_dump() const;
-  std::string get_dump_aie2() const;
 };
 
 }
--- a/src/cpp/analyzer/packets.h
+++ b/src/cpp/analyzer/packets.h
@@ -18,6 +18,7 @@ private:
 public:
   packets(const char *buffer, uint64_t size) : m_buffer(buffer), m_size(size) {}
   std::string get_dump() const;
+  std::string get_dump_aie2() const;
 };
 
 }
--- a/src/cpp/analyzer/packets.h
+++ b/src/cpp/analyzer/packets.h
@@ -12,7 +12,7 @@ class packets {
 private:
   const char *m_buffer;
   uint64_t m_size;
-  aiebu::aiebu_assembler::buffer_type m_buffer_type;
+  const aiebu::aiebu_assembler::buffer_type m_buffer_type;
 
 private:
   size_t serialize(std::ostream &stream, size_t offset) const;
--- a/src/cpp/analyzer/reporter.cpp
+++ b/src/cpp/analyzer/reporter.cpp
@@ -4,6 +4,8 @@
 #include "packets.h"
 #include "transaction.hpp"
 #include "transaction.hpp"
+#include "common/file_utils.h"
+
 
 #include "aiebu/aiebu_error.h"
 
@@ -63,9 +65,21 @@ namespace aiebu {
                     std::ofstream stream(file);
                     stream << ";  [" << i << "] " << psec->get_name() << "\t"
                            << psec->get_size() << 'B' << std::endl;
-
                     packets pprint(psec->get_data(), psec->get_size());
-                    stream << pprint.get_dump();
+                    if (is_pm_ctrlpkt(psec->get_name())) {
+                        stream << pprint.get_dump();
+                    } else if (is_ctrldata(psec->get_name())) {
+                        // Check type of control packet
+                        aiebu::aiebu_assembler::buffer_type packet_type =
+                        check_control_packet(psec->get_data(), psec->get_size());
+                        if (packet_type == aiebu::aiebu_assembler::buffer_type::blob_control_packet) {
+                            stream << pprint.get_dump();
+                        } else if (packet_type == aiebu::aiebu_assembler::buffer_type::blob_control_packet_aie2) {
+                            stream << pprint.get_dump_aie2();
+                        } else {
+                            stream << "\nInvalid control packet type" << std::endl;
+                        }
+                    }
                 }
                 continue;
             }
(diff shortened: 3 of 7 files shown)


src/cpp/analyzer/packets.cpp
```diff
@@ -49,4 +49,16 @@ packets::get_dump() const
   return stream.str();
 }
 
+std::string
+packets::get_dump_aie2() const
```
sonals: Do not add a new function for aie2. Can you add a third  parameter aiebu::aiebu_assembler::buffer_type buffer_type to the packets constructor? packets::get_dump() can then use that field to determine offset calculation.

src/cpp/analyzer/packets.h
```diff
@@ -18,6 +18,7 @@ class packets {
 public:
   packets(const char *buffer, uint64_t size) : m_buffer(buffer), m_size(size) {}
   std::string get_dump() const;
+  std::string get_dump_aie2() const;
```
sonals: As mentioned above, update packets constructor and drop get_dump_aie2() function.

src/cpp/analyzer/reporter.cpp
```diff
@@ -63,9 +65,21 @@ namespace aiebu {
                     std::ofstream stream(file);
                     stream << ";  [" << i << "] " << psec->get_name() << "\t"
                            << psec->get_size() << 'B' << std::endl;
-
                     packets pprint(psec->get_data(), psec->get_size());
-                    stream << pprint.get_dump();
+                    if (is_pm_ctrlpkt(psec->get_name())) {
```
sonals: You can remove the conditional code. Pass the buffer_type to packets constructor above.

src/cpp/analyzer/reporter.cpp
```diff
@@ -97,7 +111,20 @@ namespace aiebu {
                     stream << "Section[" << i << "]: " << psec->get_name()
                            << "\tSize: " << psec->get_size() << 'B' << std::endl;
                     packets pprint(psec->get_data(), psec->get_size());
-                    stream << "\n" << pprint.get_dump();
+                    if (is_pm_ctrlpkt(psec->get_name())) {
```
sonals: Same suggestion as above.

src/cpp/common/file_utils.cpp
```diff
@@ -40,18 +40,57 @@ identify_buffer_type(const std::vector<char>& buffer)
...
+}
+
+aiebu_assembler::buffer_type
+check_control_packet(const char* buffer, uint64_t size)
+{
+  if (size < magic_length)
+    return aiebu_assembler::buffer_type::unspecified;
+
+  auto control_packet = reinterpret_cast<const unsigned int *>(buffer);
+  uint64_t offset = 0;
```
sonals: Move the offset declaration to the line where it is used below.

src/cpp/common/file_utils.h
```diff
@@ -39,6 +39,12 @@ struct cp_ctrlinfo_aie2p
   uint32_t parity : 1;
 };
 
+// Size of word in bytes
+constexpr unsigned int word_size = 4;
+
+// For AIE2 control packets are 8 words (8words * 4bytes/word = 32bytes) aligned
+constexpr unsigned int ctrlpkt_offset_aie2 = 8 * word_size;
```
sonals: Do we need these two definitions any where else? If not move them to file_utis.cpp

src/cpp/common/file_utils.cpp
```diff
@@ -40,18 +40,57 @@ identify_buffer_type(const std::vector<char>& buffer)
...
+        odd_parity_check(*control_packet) && odd_parity_check(*(control_packet + 1))) {
+      offset = ((sizeof(*cphdr) + sizeof(*cp)) / word_size) + (cp->num_data_beat + 1);
+      control_packet += offset;
+      count += offset * word_size;
+    } else {
+      break;
+    }
+  }
+
+  if (count >= size)
```
sonals: Can you have `count > size`?

sonals: Also, here we are scanning the entire ctrlpkt buffer which may be expensive. Are we doing it even for cases where we know the platform type (like aie2p, aie2ps, etc.) from aiebu command line? Is there any other pattern in the binary we can use to distinguish between standard ctrlpkt layout and the special aie2 ctrlpkt layout? e.g., are the holes in aie2 always zero?

src/cpp/analyzer/packets.cpp
donwalkarsoham: ok sure, made the change in the latest commit.

src/cpp/analyzer/packets.h
donwalkarsoham: Yes, made the change in the latest commit

src/cpp/analyzer/reporter.cpp
donwalkarsoham: done

src/cpp/analyzer/reporter.cpp
donwalkarsoham: done

src/cpp/common/file_utils.h
donwalkarsoham: Yes. as we are using ctrlpkt_offset_aie2  in the packets.cpp hence added it in the header file here.

src/cpp/common/file_utils.cpp
donwalkarsoham: Updated to `count ==  size`.

Scanning the entire ctrlpkt buffer -
Currently the odd parity and reserved bit condition we are using to check the control packet is a necessary but not a sufficient condition. 
Hence, by checking all the control packets in the binary, at least we increase the chances of detecting the buffer being a control packet one. But this is still not a sufficient condition, and we will need to add more checks going forward like checking if value of num_data_beat <=4 for AIE2 and AIE2P control packets etc. So, as we don't have definite detectors for a binary/buffer being control packet hence, we are scanning all the control packet entries.

AIE2 v/s ASIE2P-
In this implementation the user doesn't need to know the platform as the tool detects the platform on its own.
Moreover, "-m" option in dump tool uses currently used architecture by aiebu-asm which is "(aie2ps/aie2asm/aie2txn/aie2dpu)". Here we don't have differentiate between aie2 and aie2p.
I checked with the compiler team and they just zero out the padding in case of aie2.

src/cpp/analyzer/packets.h
```diff
@@ -4,19 +4,23 @@
...
 #include <iostream>
+#include "aiebu/aiebu_assembler.h"
 
 namespace aiebu {
 
 class packets {
 private:
   const char *m_buffer;
   uint64_t m_size;
+  aiebu::aiebu_assembler::buffer_type m_buffer_type;
```
sonals: `const`

src/cpp/common/file_utils.h
sonals: Please move these definitions to packets.cpp if it is the only user.

src/cpp/common/file_utils.cpp
```diff
@@ -40,18 +40,56 @@ identify_buffer_type(const std::vector<char>& buffer)
...
-  if ((cphdr->reserved_a_0 == 0x0) && (cphdr->reserved_b_0 == 0x0) &&
-      (cphdr->reserved_c_000 == 0x0) && (cp->reserved_00 == 0x0)) {
-    if (odd_parity_check(data[0]) && odd_parity_check(data[1]))
-      return aiebu_assembler::buffer_type::blob_control_packet;
-    else
+  return check_control_packet(buffer.data(), buffer.size());
+}
+
+aiebu_assembler::buffer_type
+check_control_packet(const char* buffer, uint64_t size)
```
sonals: Rename to `identify_control_packet`

src/cpp/common/file_utils.h
donwalkarsoham: packets.cpp and file_utils.cpp are the only two users.
Added to both separately.

sonals (approved):

PR #98, opened 2025-06-11
Add light weight AIEBU User Manual

donwalkarsoham:
Problem solved by the commit
Add AIEBU User Manual as per below guideline

Add example usage of aiebu-asm command leaving out the other use cases of aiebu-asm.
Example ELF file layout.
Add an example usage of aiebu-dump
Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
N/A

How problem was solved, alternative solutions (if any) and why they were rejected
Created a .rst file having above details.

Risks (if any) associated the changes in the commit
N/A

What has been tested and how, request additional testing if necessary
This is generated from https://amd.atlassian.net/wiki/spaces/AIE/pages/916173309/AIEBU+Usage+Manual+In+progress.
This page can be used to verify the user manual as the format and contents is exactly the same

Documentation impact (if any)
This is addition of new .rst document to the repo

### DIFF
--- a/AIEBU_USER_MANUAL.rst
+++ b/AIEBU_USER_MANUAL.rst
@@ -1,0 +1,387 @@
+AIEBU Usage Manual (In progress)
+================================
+
+-  `AIE Binary Utilities
+   (AIEBU) <#AIEBUUsageManual(Inprogress)-AIEBinaryU>`__
+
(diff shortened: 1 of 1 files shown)


sonals:
It needs a bit more cleanup, I will make that change on top of this.

PR #99, opened 2025-06-12
Add Support for Txn Ctrlcode binary to asm conversion to Dump tool

donwalkarsoham:
Support for Txn Ctrlcode binary to asm conversion to Dump tool

N/A

Currently added new APIs for blob dump in reporter class to not break the existing disassemble(), ctrlcode_summary() which are used by AIEBU assembler.
1. -d and -D option will print the control code .asm file from .bin file
2. -p will display the frequency of each opcode present in the control code binary.

N/A

1> Printed control codes using -d and -D option by providing .bin files and .elf files and compared the output.
2> Printed output for -f option to see opcode frequency.

1. Opcode frequency
**./aiebu-dump -p <path>/<ctrlcode.bin>**

aie2-ctrlcode
v0.1, gen4
6x4 M1
435188B, 13789ops
XAIE_IO_WRITE                               9048
XAIE_IO_BLOCKWRITE                          3232
XAIE_IO_MASKWRITE                              0
XAIE_IO_MASKPOLL                               0
XAIE_IO_MASKPOLL_BUSY                          0
XAIE_IO_NOOP                                   0
XAIE_IO_PREEMPT                               35
XAIE_IO_LOAD_PM_START                          1
XAIE_IO_CUSTOM_OP_TCT                         59
XAIE_IO_CUSTOM_OP_DDR_PATCH                 1220
XAIE_IO_CUSTOM_OP_MERGE_SYNC                 194
XAIE_IO_CUSTOM_OP_RECORD_TIMER                 0

2. ASM Control code from binary using -d option.
**./aiebu-dump -d <path>/<ctrlcode.bin >**

    .attach_to_group 0
    .attach_to_group 1
    .attach_to_group 2
    .attach_to_group 3

XAIE_IO_LOAD_PM_START           0x9, #2
XAIE_IO_BLOCKWRITE              @0x1d000, [8]
XAIE_IO_BLOCKWRITE.0            0x1291
XAIE_IO_BLOCKWRITE.1            0xbb88
XAIE_IO_BLOCKWRITE.2            0x0
XAIE_IO_BLOCKWRITE.3            0x0
XAIE_IO_BLOCKWRITE.4            0xc0000000
XAIE_IO_BLOCKWRITE.5            0x2000000
XAIE_IO_BLOCKWRITE.6            0x0
XAIE_IO_BLOCKWRITE.7            0x2000000
XAIE_IO_WRITE                   @0x1d214, 0x80000000
...

3. ASM Control code from binary and opcode frequency using -D option.
**./aiebu-dump -D <path>/<ctrlcode.bin>**

    .attach_to_group 0
    .attach_to_group 1
    .attach_to_group 2
    .attach_to_group 3

XAIE_IO_LOAD_PM_START           0x9, #2
XAIE_IO_BLOCKWRITE              @0x1d000, [8]
XAIE_IO_BLOCKWRITE.0            0x1291
XAIE_IO_BLOCKWRITE.1            0xbb88
XAIE_IO_BLOCKWRITE.2            0x0
XAIE_IO_BLOCKWRITE.3            0x0
XAIE_IO_BLOCKWRITE.4            0xc0000000
XAIE_IO_BLOCKWRITE.5            0x2000000
XAIE_IO_BLOCKWRITE.6            0x0
XAIE_IO_BLOCKWRITE.7            0x2000000
XAIE_IO_WRITE                   @0x1d214, 0x80000000
...

v0.1, gen4
6x4 M1
435188B, 13789ops
XAIE_IO_WRITE                               9048
XAIE_IO_BLOCKWRITE                          3232
XAIE_IO_MASKWRITE                              0
XAIE_IO_MASKPOLL                               0
XAIE_IO_MASKPOLL_BUSY                          0
XAIE_IO_NOOP                                   0
XAIE_IO_PREEMPT                               35
XAIE_IO_LOAD_PM_START                          1
XAIE_IO_CUSTOM_OP_TCT                         59
XAIE_IO_CUSTOM_OP_DDR_PATCH                 1220
XAIE_IO_CUSTOM_OP_MERGE_SYNC                 194
XAIE_IO_CUSTOM_OP_RECORD_TIMER                 0

N/A

### DIFF
--- a/src/cpp/analyzer/reporter.cpp
+++ b/src/cpp/analyzer/reporter.cpp
@@ -6,8 +6,6 @@
 #include "transaction.hpp"
 #include "common/file_utils.h"
 
-
-
 #include "aiebu/aiebu_error.h"
 
 #include <boost/interprocess/streams/bufferstream.hpp>
--- a/src/cpp/analyzer/reporter.cpp
+++ b/src/cpp/analyzer/reporter.cpp
@@ -203,8 +203,7 @@ namespace aiebu {
     void reporter::disassemble_blob(const std::filesystem::path &root) const
     {
         std::filesystem::path file(root);
-        file += "disassemble_";
-        file += "blob.asm";
+        file += "_disassemble.asm";
         std::ofstream stream(file);
         if (!stream) {
             throw error(error::error_code::internal_error, "Failed to open file for writing");
--- a/src/cpp/analyzer/reporter.cpp
+++ b/src/cpp/analyzer/reporter.cpp
@@ -15,15 +15,13 @@
 
 namespace aiebu {
 
-reporter::reporter(aiebu::aiebu_assembler::buffer_type type, const std::vector<char>& buffer)
-: m_buffer_type(type), m_buffer(buffer)
+reporter::reporter(aiebu::aiebu_assembler::buffer_type type, const std::vector<char>& elf_data) 
+    : m_buffer_type(type), m_buffer(elf_data.data()), m_buffer_size(elf_data.size())
 {
-    boost::interprocess::ibufferstream istr(buffer.data(), buffer.size());
-    if (type == aiebu::aiebu_assembler::buffer_type::elf_aie2) {
-        bool result = my_elf_reader.load(istr);
-        if (!result) {
-            throw error(error::error_code::invalid_buffer_type, "Invalid ELF buffer");
-        }
+    boost::interprocess::ibufferstream istr(elf_data.data(), elf_data.size());
+    bool result = my_elf_reader.load(istr);
+    if (!result) {
+        throw error(error::error_code::invalid_buffer_type, "Invalid ELF buffer");
     }
 }
 
@@ -126,13 +124,13 @@ void reporter::disassemble(std::ostream &stream, bool all) const
 
 void reporter::ctrlcode_blob_summary(std::ostream &stream) const
 {
-    transaction tprint(m_buffer.data(), m_buffer.size());
+    transaction tprint(m_buffer, m_buffer_size);
     stream << tprint.get_txn_summary() << std::endl;
 }
 
 void reporter::disassemble_blob(std::ostream &stream) const
 {
-    transaction tprint(m_buffer.data(), m_buffer.size());
+    transaction tprint(m_buffer, m_buffer_size);
     stream << tprint.get_all_ops() << std::endl;
 }
 
--- a/src/cpp/analyzer/reporter.cpp
+++ b/src/cpp/analyzer/reporter.cpp
@@ -6,130 +6,172 @@
 #include "transaction.hpp"
 #include "common/file_utils.h"
 
+
 #include "aiebu/aiebu_error.h"
 
 #include <boost/interprocess/streams/bufferstream.hpp>
 #include <elfio/elfio_dump.hpp>
-
 namespace aiebu {
 
-reporter::reporter(aiebu::aiebu_assembler::buffer_type type, const std::vector<char>& elf_data) 
-    : m_buffer_type(type), m_buffer(elf_data.data()), m_buffer_size(elf_data.size())
-{
-    boost::interprocess::ibufferstream istr(elf_data.data(), elf_data.size());
-    bool result = my_elf_reader.load(istr);
-    if (!result) {
-        throw error(error::error_code::invalid_buffer_type, "Invalid ELF buffer");
+    reporter::reporter(aiebu::aiebu_assembler::buffer_type type, const std::vector<char>& buffer) :
+        m_buffer_type(type), m_buffer(buffer)
+    {
+        if (type == aiebu::aiebu_assembler::buffer_type::elf_aie2) {
+            boost::interprocess::ibufferstream istr(buffer.data(), buffer.size());
+            bool result = my_elf_reader.load(istr);
+            if (!result)
+                throw error(error::error_code::invalid_buffer_type, "Invalid ELF buffer");
+        }
     }
-}
 
-void reporter::elf_summary(std::ostream &stream) const
-{
-    ELFIO::dump::header(stream, my_elf_reader);
-    ELFIO::dump::section_headers(stream, my_elf_reader);
-    ELFIO::dump::segment_headers(stream, my_elf_reader);
-}
-
-void reporter::ctrlcode_summary(std::ostream &stream) const
-{
-    ELFIO::Elf_Half sec_num = my_elf_reader.sections.size();
-    for (int i = 0; i < sec_num; ++i) {
-        const ELFIO::section* psec = my_elf_reader.sections[i];
-
-        // Decoding not supported for ".ctrldata" section
-        // for aie2 ".ctrldata" contain control packet and ".ctrlpkt-pm-N" contain
-        // pm control packet which cannot be decoded
-        if (psec->get_type() != ELFIO::SHT_PROGBITS || is_ctrldata(psec->get_name())
-           || is_pm_ctrlpkt(psec->get_name()))
-            continue;
-
-        stream << "  [" << i << "] " << psec->get_name() << "\t"
-               << psec->get_size() << std::endl;
-
-        transaction tprint(psec->get_data(), psec->get_size());
-        stream << tprint.get_txn_summary() << std::endl;
+    void reporter::elf_summary(std::ostream &stream) const
+    {
+        if (m_buffer_type != aiebu::aiebu_assembler::buffer_type::elf_aie2) {
+            throw error(error::error_code::invalid_buffer_type, "Invalid buffer type for ELF summary");
+        }
+        ELFIO::dump::header(stream, my_elf_reader );
+        ELFIO::dump::section_headers( stream, my_elf_reader);
+        ELFIO::dump::segment_headers( stream, my_elf_reader);
     }
-}
 
-void reporter::disassemble(const std::filesystem::path &root, bool all) const
-{
-    ELFIO::Elf_Half sec_num = my_elf_reader.sections.size();
-    for (int i = 0; i < sec_num; ++i) {
-        const ELFIO::section* psec = my_elf_reader.sections[i];
+    void reporter::ctrlcode_summary(std::ostream &stream) const
+    {
+        if (m_buffer_type == aiebu::aiebu_assembler::buffer_type::elf_aie2) {
+            ELFIO::Elf_Half sec_num = my_elf_reader.sections.size();
+            for ( int i = 0; i < sec_num; ++i ) {
+                const ELFIO::section* psec = my_elf_reader.sections[i];
 
-        if (psec->get_type() != ELFIO::SHT_PROGBITS)
-            continue;
-        if (is_pm_ctrlpkt(psec->get_name()) || is_ctrldata(psec->get_name())) {
-            if (all) {
+                // Decoding not supported for ".ctrldata" section
+                // for aie2 ".ctrldata" contain control packet and ".ctrlpkt-pm-N" contain
+                // pm control packet which cannot be decoded
+                if (psec->get_type() != ELFIO::SHT_PROGBITS || is_ctrldata(psec->get_name())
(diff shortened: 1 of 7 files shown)


sonals:
We should strive to keep aiebu-dump behavior like GNU objdump. Let's not tie -D to the summary dumping. For ctrlcode, switches -d and -D should do the same. To display ctrlcode summary information we can use **-p**? See the GNU objdump usage:

`-p, --private-headers    Display object format specific file header contents`

donwalkarsoham:
> We should strive to keep aiebu-dump behavior like GNU objdump. Let's not tie -D to the summary dumping. For ctrlcode, switches -d and -D should do the same. To display ctrlcode summary information we can use **-p**? See the GNU objdump usage:
> 
> `-p, --private-headers Display object format specific file header contents`



> We should strive to keep aiebu-dump behavior like GNU objdump. Let's not tie -D to the summary dumping. For ctrlcode, switches -d and -D should do the same. To display ctrlcode summary information we can use **-p**? See the GNU objdump usage:
> 
> `-p, --private-headers Display object format specific file header contents`

donwalkarsoham:
> We should strive to keep aiebu-dump behavior like GNU objdump. Let's not tie -D to the summary dumping. For ctrlcode, switches -d and -D should do the same. To display ctrlcode summary information we can use **-p**? See the GNU objdump usage:
> 
> `-p, --private-headers Display object format specific file header contents`



> We should strive to keep aiebu-dump behavior like GNU objdump. Let's not tie -D to the summary dumping. For ctrlcode, switches -d and -D should do the same. To display ctrlcode summary information we can use **-p**? See the GNU objdump usage:
> 
> `-p, --private-headers Display object format specific file header contents`

Sure, I will make output for -d and -D options same for control code binary. For display of only ctrlcode summary which is actually the opcode frequency I was planning to use -f "Display the contents of the overall file header" option (I think I missed to submit the corresponding dump.cpp, as if you see the commit message it has the -f option but it doesn't reflect in the source code in commit). But yes, I can add the -p option for the control code summary.

donwalkarsoham:
retest this please

src/cpp/analyzer/reporter.cpp
```diff
@@ -7,114 +7,131 @@
...
+
+        stream << "\n";
+        stream << "Section[" << i << "]: " << psec->get_name() << "\tSize: "
+               << psec->get_size() << 'B' << std::endl;
+        transaction tprint(psec->get_data(), psec->get_size());
+        stream << tprint.get_all_ops() << std::endl;
     }
 }
+
+void reporter::ctrlcode_blob_summary(std::ostream &stream) const
```
sonals: Please make this method _private_ and call it from `reporter::ctrlcode_summary() when buffer type is blob`.  The clients of reporter class  should not need call different methods for different buffer type.

src/cpp/analyzer/reporter.cpp
```diff
@@ -7,114 +7,131 @@
...
+{
+    transaction tprint(m_buffer, m_buffer_size);
+    stream << tprint.get_txn_summary() << std::endl;
+}
+
+void reporter::disassemble_blob(std::ostream &stream) const
+{
```
sonals: Same comment here.

src/cpp/analyzer/reporter.h
```diff
@@ -9,26 +9,33 @@
...
-        reporter(aiebu::aiebu_assembler::buffer_type type, const std::vector<char>& elf_data);
-        void elf_summary(std::ostream &stream) const;
-        void ctrlcode_summary(std::ostream &stream) const;
-        void disassemble(const std::filesystem::path &root, bool all = false) const;
-        void disassemble(std::ostream &stream, bool all = false) const;
-    };
+  class reporter {
+  private:
+      ELFIO::elfio my_elf_reader;
+      const aiebu::aiebu_assembler::buffer_type m_buffer_type;
```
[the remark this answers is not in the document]
sonals: One way to turn off this warning would to mark copy and move operators with delete. Look up examples on how to do this.

src/cpp/analyzer/reporter.h
```diff
@@ -9,26 +9,34 @@
...
+  public:
+      reporter(aiebu::aiebu_assembler::buffer_type type, const std::vector<char>& elf_data);
+      reporter(aiebu::aiebu_assembler::buffer_type type, const char* buffer, uint64_t size) :  m_buffer_type(type), m_buffer(buffer), m_buffer_size(size) {}
+      void elf_summary(std::ostream &stream) const;
+      void ctrlcode_summary(std::ostream &stream) const;
+      void disassemble(const std::filesystem::path &root, bool all = false) const;
+      void disassemble(std::ostream &stream, bool all = false) const;
+      void ctrlcode_blob_summary(std::ostream &stream) const;
+      void disassemble_blob(std::ostream &stream) const;
+  };
```
sonals: Remove these two public methods.

src/cpp/utils/dump/dump.cpp
```diff
@@ -119,6 +119,8 @@ int main(int argc, char* argv[])
   aiebu::aiebu_assembler::buffer_type type = aiebu::identify_buffer_type(buffer);
 
   std::cout << aiebu::buffer_type_table.at(type) << std::endl;
+
+
```
sonals: Why additional line here?

src/cpp/utils/dump/dump.cpp
```diff
@@ -139,5 +141,14 @@ int main(int argc, char* argv[])
...
   }
+  else if (type == aiebu::aiebu_assembler::buffer_type::blob_instr_transaction) {
+    aiebu::reporter rep(type, buffer.data(), buffer.size());
+    if (result["all-headers"].as<bool>()) {
+      rep.ctrlcode_blob_summary(std::cout);
+    }
+    else if (result["disassemble"].as<bool>()) {
+      rep.disassemble_blob(std::cout);
+    }
+  }
```
sonals: All this conditional code can go away once you make the suggested changes to reporter class.

sonals (changes requested):
Please make the suggested changes.

src/cpp/analyzer/reporter.cpp
donwalkarsoham: Yes, I had made similar change (have it locally) but that breaks the existing implementation. Hence, I have added separate APIs for blob to maintain backward compatibility.

This is because the aiebu_assembler.cpp assumes that the reporter class only processes ELF types and hence when it converts binary to ELF and then call reporter class APIs to dump the ELF it doesn't change the buffer type from elf_aie2 to blob_instr_transaction before calling reporter APIs like reporter::ctrlcode_summary(std::ostream &stream) and reporter::disassemble(std::ostream &stream). Below is the code for reference, here m_type remains unchanged.

Code reference - https://github.com/Xilinx/aiebu/blob/main-ge/src/cpp/assembler/aiebu_assembler.cpp#L84

I can update get_report() or other APIs here to change buffer type and then call reporter::ctrlcode_summary(), but the issue is that if any other customer has used aiebu_assembler() and get_report() in a similar manner then it might break those customers.
Hence, to maintain backward compatibility I have added new APIs for blob.

src/cpp/utils/dump/dump.cpp
donwalkarsoham: will remove this.

src/cpp/analyzer/reporter.cpp
donwalkarsoham: same as above

src/cpp/analyzer/reporter.h
donwalkarsoham: same as above.

src/cpp/analyzer/reporter.h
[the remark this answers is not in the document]
donwalkarsoham: done

src/cpp/analyzer/reporter.cpp
donwalkarsoham: I have also made your suggested changes in the below commit. This has updates to the aiebu_assembler.cpp API implementation though. Please let me know if you want me to use the code changes in this commit.
https://github.com/donwalkarsoham/aiebu/commit/e409a2f20a24204d490e7d9f20af48f0d42275e5

sonals: I do not see why any public API change is required for `aiebu_assembler` class. The constructor already requires clients to specify the `buffer_type`. Are you referring to the C API? Then we should extend `enum aiebu_assembler_buffer_type` to match `enum class buffer_type`.

Essentially the reporter should do the _right_ thing given a buffer since it already knows about the buffer_type which was passed by aiebu_assembler.

donwalkarsoham: Yes, the user provides the buffer type which is corresponding to the input binary or asm files.
When this is given to the `aiebu_assembler` constructor it converts these input binary/asm files to ELF files. 
But the constructor doesn't change the buffer_type to `aie2_elf` from `blob_instr_transaction`, for example. 
Reference - https://github.com/Xilinx/aiebu/blob/main-ge/src/cpp/assembler/aiebu_assembler.cpp#L47

Now the `aiebu_assembler::get_report()` API passes the same buffer_type to the reporter Class without changing the buffer_type to one of the ELF buffer types which is incorrect as ideally the buffer_type should be updated to ELF type as aiebu_assembler converts the buffer from bin/asm to ELF.
Reference - https://github.com/Xilinx/aiebu/blob/main-ge/src/cpp/assembler/aiebu_assembler.cpp#L86

Hence, adding buffer_type conditional code in existing `reporter::disassemble()` API fails the existing tests.

sonals: There are few enhancements that should be made. 
1. `aiebu_assembler `should have an additional field for output buffer_type; call it **m_ouput_type**. m_type can remain the input buffer type as today.
2. `reporter` should _not_ throw away the `type` information in the constructor. 
3. `reporter` should use the type information to support printing of all types of buffers, not just elf.

donwalkarsoham: Sure, thanks. I had created below commit has #2 and #3 but its using the existing m_type buffer to avoid adding a new parameter. But will make the suggested changes to it and add m_output_type.
https://github.com/donwalkarsoham/aiebu/commit/e409a2f20a24204d490e7d9f20af48f0d42275e5

src/cpp/utils/dump/dump.cpp
donwalkarsoham: done

donwalkarsoham:
> Please make the suggested changes.

done

src/cpp/utils/dump/dump.cpp
```diff
@@ -117,27 +118,33 @@ int main(int argc, char* argv[])
...
+      }
+      else if (result["disassemble-all"].as<bool>()) {
+        if (type == aiebu::aiebu_assembler::buffer_type::elf_aie2) {
+          rep.disassemble(std::cout, true);
+        }
+      }
+      else if (result["private-headers"].as<bool>()) {
+        rep.ctrlcode_summary(std::cout);
+      }
+    }
```
sonals, line 148: It would be best to refactor the command line parsing code to index off the command line switch rather than the buffer type. Let's go with this for now, we should work on cleanup later.

sonals (approved):

src/cpp/utils/dump/dump.cpp
donwalkarsoham: Got it, sure. Will update going forward.

src/cpp/assembler/aiebu_assembler.cpp
```diff
@@ -36,35 +36,44 @@ aiebu_assembler(buffer_type type,
                 const std::vector<std::string>& libpaths,
                 const std::map<uint32_t, std::vector<char> >& ctrlpkt) : m_type(type)
 {
+  if (buffer1.empty()) {
```
HimanshuChoudhary-Xilinx, line 39: we should not validate input here, this condition is not true for aie2/aie2ps/aie4 config targets causing breakage

PR #114, opened 2025-07-09
update patch_op_opt_t struct for 1.0 TXN bin

HimanshuChoudhary-Xilinx:
update patch_op_opt_t struct for 1.0 TXN bin
[PR214]: patch_op_opt_t and scratchoffset and TXN 1.0 changes


larry9523 (commented):
After this change, the save/restore TXN control code will use optimized data structure of patch op code. We need to make sure the FW change is in place to handle the optimized patch opcode instead of the 0.1 patch opcode. 

@HimanshuChoudhary-Xilinx  can you please work with Sri Latha to understand the current status of FW (TXN 1.0 control code parse) about what structure it is using to parse 1.0 TXN patch opcode. If it is still using 0.1 structure, let's fix FW first before merging this PR.

HimanshuChoudhary-Xilinx:
> After this change, the save/restore TXN control code will use optimized data structure of patch op code. We need to make sure the FW change is in place to handle the optimized patch opcode instead of the 0.1 patch opcode.
> 
> @HimanshuChoudhary-Xilinx can you please work with Sri Latha to understand the current status of FW (TXN 1.0 control code parse) about what structure it is using to parse 1.0 TXN patch opcode. If it is still using 0.1 structure, let's fix FW first before merging this PR.

hi @larry9523 
It will never fail in FW as patching info is generate in aiebu and patched in xrt. In FW it become NOP and we increment by Hdr->Size.
Nobody generate 1.0 TXN bin except save/restore.
We checked FW code, it use patch_op_opt_t (new struct) but nobody tested FM patching for 1.0TXNbin/save/restore. It would have failed.
But with this change it will pas. and can be patched in fw also.

larry9523 (approved):

PR #119, opened 2025-07-17
dynamic tracing compiler

AdvaitNaik:

CERT dynamic tracing compiler
[cert dynamic tracing support in baremetal flow (dtrace compiler)](https://confluence.amd.com/pages/viewpage.action?pageId=1665323347)

Rebasing from GHE:
PR[110]: CERT dynamic tracing compiler
PR[198]: CERT dynamic tracing mask write register action 
PR[200]: CERT dynamic tracing return error message 
PR[219]: CERT dynamic tracing map search optimization 

NA

NA

None

CERT baremetal flow

NA


xuhz (approved):

xuhz (approved):

src/cpp/CMakeLists.txt
```diff
@@ -117,3 +117,6 @@ install(FILES
   DESTINATION ${AIEBU_INSTALL_INCLUDE_DIR}
   CONFIGURATIONS Debug Release COMPONENT Runtime
 )
+
+# Build dynamic tracing library
+add_subdirectory(dtrace)
```
stsoe: Add missing newline

stsoe (changes requested):
Can you rebase please, there were changes to setings.cmake and in how AIEBU is built.  Just want to make sure your changes will still build as expected. 

Also add the missing newline

src/cpp/CMakeLists.txt
AdvaitNaik: done

test/dtrace_test/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+
+add_executable(${DTRACE_TESTNAME} dtrace_test.cpp)
+
+add_dependencies(${DTRACE_TESTNAME} cert_dtrace)
+
+if (MSVC)
+  set(DTRACE_LIBRARY_PATH "${DTRACE_SOURCE_DIR}/$<CONFIG>/cert_dtrace.dll")
+else()
+  set(DTRACE_LIBRARY_PATH "${DTRACE_SOURCE_DIR}/libcert_dtrace.so")
+endif()
```
stsoe, line 17: This is too specific to this deeply buried CMakeLists.txt file, something is needed in setttings.cmake.
I know kind of the same trick was used under test/cmake_test/CMakeLists.txt, but makefile's variable belong in settings.cmake as:
```
if (WIN32)
  # On windows the build script uses multi-config build, meaning same
  # binary dir is used for Release and Debug.  At install time,
  # artifacts are staged into Release/ and Debug/ subdirectories
  set(AIEBU_INSTALL_STAGING_DIR ${CMAKE_BINARY_DIR}/$<CONFIG>/xilinx/aiebu)
else()
  # On Linux the build script uses separate binary dir for Release or Debug
  set(AIEBU_INSTALL_STAGING_DIR ${CMAKE_BINARY_DIR}/opt/xilinx/aiebu)
endif()
```
What I don't understand is what your dll/so is doing under  `"${AIEBU_BINARY_DIR}/src/cpp/dtrace"`  Are they not installable artifacts?

AdvaitNaik: I have added dtrace configuration in settings.cmake `set(AIEBU_DTRACE_DIR "${AIEBU_BINARY_DIR}/dtrace")` for dtrace artifacts to properly staged during the installation process. 
However, for the test cases, I am using the dll/so files directly from the source directory for the following reasons-

Custom Command - The test case CMakeLists.txt file use a custom_command to create the control block using the built dtrace compiler library. 
The add_test command only then compares the MD5 checksum of the generated control block with the gold standard.

Using the install directory would require splitting each test case into two and would double the number of test cases for the dtrace compiler, which I wanted to avoid.
- One to create the control block.
- Another to compare the MD5 checksum. 

Since the install directory is only available after the make process completes, so dll/so files under `${AIEBU_BINARY_DIR}/src/cpp/dtrace` are used temporarily for the test cases custom_command during the build process.

I used this approach to ensure functionality of the dtrace compiler is tested without increasing the number of test cases unnecessarily.

stsoe (changes requested):
-- Installing: C:/Users/stsoe/git/stsoe/aiebu/build/WBuild/dtrace/cert_dtrace.lib
This directory location is wrong, link artifacts must be installed to AIEBU_INSTALL_LIB_DIR in AIEBU_DEV_COMPONENT (see `src/cpp/CMakeLlist.txt` for how)

Btw.  What good is a `.lib` file without header API files?  How is this .lib file used?

AdvaitNaik:
> -- Installing: C:/Users/stsoe/git/stsoe/aiebu/build/WBuild/dtrace/cert_dtrace.lib This directory location is wrong, link artifacts must be installed to AIEBU_INSTALL_LIB_DIR in AIEBU_DEV_COMPONENT (see `src/cpp/CMakeLlist.txt` for how)
> 
> Btw. What good is a `.lib` file without header API files? How is this .lib file used?

My bad. Corrected files installed and also installed directory location with correct component

src/cpp/dtrace/action/action_control.h
```diff
@@ -0,0 +1,570 @@
...
+#ifndef ACTION_CONTROL_H
+#define ACTION_CONTROL_H
+
+// This file contains the declaration of the action control class and action classes.
+#include <vector>
+#include <string>
+#include <map>
+#include <unordered_map>
+#include <cstdint>
+#include <regex>
```
stsoe: Includes should be sorted in groups.  I typically do in three groups:
```
<local includes>

<boost or other 3rd party includes>

<system includes>
```

src/cpp/dtrace/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+  )
+
+add_library(cert_dtrace SHARED
+  $<TARGET_OBJECTS:dtrace_library_objects>
+  )
+
+install(TARGETS cert_dtrace
+  LIBRARY DESTINATION ${AIEBU_INSTALL_LIB_DIR}/dtrace COMPONENT ${AIEBU_DEV_COMPONENT}
+  RUNTIME DESTINATION ${AIEBU_INSTALL_LIB_DIR}/dtrace COMPONENT ${AIEBU_DEV_COMPONENT}
+  )
```
stsoe, line 50: We are going to run into upstreaming conflicts when using /usr/lib/dtrace/, it will not be allowed.   Also any libraries released must be versioned shared objects.  

I really have to understand how cert_dtrace or any artifacts related by by src/cpp/dtrace are expected to be used and by whom?  I have probably asked this before and apologize for repeating myself.   

The PR cannot be merged before we have resolved the artifacts released.

stsoe (changes requested):

src/cpp/dtrace/CMakeLists.txt
AdvaitNaik: adding comment for reference that we discussed offline

USER CONFIGURATION
https://github.com/Xilinx/XRT/blob/f47d3cd549dff89f7bcaa48b83254c40f4598503/src/runtime_src/core/common/config_reader.h#L1074
Users specify the dtrace library path and control file path in the xrt.ini
[Debug]
dtrace_lib_path=/path/to/libcert_dtrace.so
dtrace_control_file_path=/path/to/script.ct

XRT INTEGRATION
https://github.com/Xilinx/XRT/blob/f47d3cd549dff89f7bcaa48b83254c40f4598503/src/runtime_src/core/common/api/xrt_module.cpp#L2398
XRT dynamically loads the dtrace library at runtime using dlopen from the path

src/cpp/dtrace/action/action_control.h
AdvaitNaik: done

sonals (commented):
Please break this large PR into smaller PRs which build on top of the previous one.

sonals:
> Please break this large PR into smaller PRs which build on top of the previous one.

@AdvaitNaik explained this code is actually being migrated from another repository where it had already been reviewed. So, no need to break this into smaller pieces.

cmake/settings.cmake
```diff
@@ -80,12 +80,12 @@ else()
...
-set(AIEBU_INSTALL_CMAKE_DIR         ${AIEBU_INSTALL_DIR}/share/cmake/${PROJECT_NAME})
-set(AIEBU_PYTHON_INSTALL_DIR        ${AIEBU_INSTALL_DIR}/${CMAKE_INSTALL_LIBDIR}/python3)
-set(AIEBU_SPECIFICATION_INSTALL_DIR ${AIEBU_INSTALL_DIR}/share/specification)
-set(AIEBU_GEN_DIR                   ${AIEBU_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR}/gen)
+set(AIEBU_INSTALL_LIB_DIR           "${AIEBU_INSTALL_DIR}/lib")
+set(AIEBU_INSTALL_INCLUDE_DIR       "${AIEBU_INSTALL_DIR}/include")
+set(AIEBU_INSTALL_CMAKE_DIR         "${AIEBU_INSTALL_DIR}/share/cmake/${PROJECT_NAME}")
+set(AIEBU_PYTHON_INSTALL_DIR        "${AIEBU_INSTALL_DIR}/lib/python3")
+set(AIEBU_SPECIFICATION_INSTALL_DIR "${AIEBU_INSTALL_DIR}/share/specification")
+set(AIEBU_GEN_DIR                   "${AIEBU_BINARY_DIR}/lib/gen")
```
stsoe: Please revert these changes.

stsoe (changes requested):
settings.cmake should not be changed.

cmake/settings.cmake
AdvaitNaik: done

src/cpp/dtrace/utils.h
```diff
@@ -0,0 +1,105 @@
...
+    dtrace_warning = 2,
+    dtrace_info = 3
+};
+
+// Logging Prefix
+#define DTRACE_PREFIX "[DTRACE] "
+
+// Error Logging
+#ifndef DTRACE_ERROR
+#define DTRACE_ERROR(rc, fmt, ...) \
```
stsoe, line 30: Why do all macros in this file use do/while? 
Have you considered templatizing this and writing is a C++ code?
I am not pushing for this to be changed now, I am just curious why we are using macro for more that just the injection of __func__, and __line__ ?

stsoe: You could ask chatgpt or perplexity how the do/while macro could be improved to be more C++ish.  It would give you a good variadic template answer, I am sure.

AdvaitNaik: This was the initial version of logging support that I introduced in the dtrace compiler. For this PR, I wanted to port the existing gitenterprise code to the github, so I didn't made any changes to the logging macros themselves. I definitely plan to improve the logging when I work on updating the dtrace compiler paging update

stsoe (approved):

PR #122, opened 2025-07-28
Disable findpackage.cmake if cmake < 3.20

stsoe:
cmake_path used in findpackage.cmake is available only in cmake >= 3.20.  XRT builds with an insane old version of cmake on CentOS7.9


stsoe:
Temporary glitch on github.com actions.  Bypassing check of this PR.

PR #132, opened 2025-08-20
AIE4/AIE2PS: remove patching instructions when patching a buffer from host

sayyanna:
https://jira.xilinx.com/browse/AIESW-8856
Enhance aiebu to remove patching instructions when patching a buffer from host
This is an enhancement. Details are here: https://jira.xilinx.com/browse/AIESW-8856
Removed apply_patch_57 as part of text section
Disassembler may not work. This needs to be addressed
Tested the changes for aie4 on CERT simnow baremetal


HimanshuChoudhary-Xilinx:
we didnt remove Apply_offset_57 in past because we wanted to give runtime a option to decide if we want to in host or firmware. https://amd.atlassian.net/wiki/spaces/AIE/pages/779179499/Unify+host+address+patching+approaches+at+control+code+level
with this pr we are fixing at aiebu level only

sonals:
Wondering if we should add a std::map table for opcodes that should be removed? That will make this solution more generic. In addition we should have a flag that disables the optimization. Time to add -O0 to -O4 like GNU as?

src/cpp/common/assembler_state.h
```diff
@@ -223,6 +226,19 @@ class assembler_state : public std::enable_shared_from_this<assembler_state>
     return m_dependent_labelmap;
   }
 
+  void process_optimization(uint32_t optimize_level)
+  {
+    if (optimize_level >= 1) {
```
sonals: You can drop **{}** if you have only statement inside. That makes the code more succinct.

src/cpp/common/assembler_state.h
```diff
@@ -223,6 +226,19 @@ class assembler_state : public std::enable_shared_from_this<assembler_state>
...
+      m_opt_opcodes.insert("apply_offset_57");
+    } else {
+      m_opt_opcodes.clear();
+    }
+  }
+
+  bool is_enable_opt_opcode(std::string name) {
```
sonals: ` bool is_enable_opt_opcode(const std::string &name) const {`

sonals (approved):

larry9523 (approved):
Looks good. Please update the aiebu user manual once merged. 
https://amd.atlassian.net/wiki/spaces/AIE/pages/832142269/AIEBU+User+Manual

PR #138, opened 2025-09-03
Robustness improvements

sonals:

1. Improve the handling of non compliant aie2 binaries. The bugs were discovered when adding the full ELF support.
2. On Linux hide, all symbols except explicitly exported ones

NA

Check if a map or a vector is empty before trying to use its first element

None

ctest validated

NA

### DIFF
--- a/src/cpp/elf/aie2/aie2_blob_elfwriter.h
+++ b/src/cpp/elf/aie2/aie2_blob_elfwriter.h
@@ -34,8 +34,12 @@ public:
    * It add partition info in note section and finalize to generate elf
    */
   std::vector<char>
-  process(std::vector<std::shared_ptr<writer>>& mwriter) override
+  process(std::vector<std::shared_ptr<writer>> &mwriter) override
   {
+
+    if (mwriter.empty())
+      return {};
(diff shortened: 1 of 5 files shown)


HimanshuChoudhary-Xilinx (approved):

src/cpp/elf/aie2/aie2_blob_elfwriter.h
```diff
@@ -34,8 +34,12 @@ class aie2_config_elf_writer: public elf_writer
    * It add partition info in note section and finalize to generate elf
    */
   std::vector<char>
-  process(std::vector<std::shared_ptr<writer>>& mwriter) override
+  process(std::vector<std::shared_ptr<writer>> &mwriter) override
```
stsoe: Uh oh, I don't like this, it's uncommon in C++to have space before `*` and `&`, I favor Stroustrup's style.

If you really want to make changes, then fix all the functions that pass std::shared_ptr by value to pass by const ref, unless they move the value argument.

stsoe (approved):
One very minor nit.

PR #140, opened 2025-09-10
Fix clang-tidy issues and enable some aie2ps testing on Windows

sonals:

1. Remove duplicate `asm_preprocessor_input:get_keys()` as it is already present in the base class, `preprocessor_input`
2. Remove `asm_preprocessor_input:get_data()` as it was masking the virtual function `preprocessor_input::get_data(const std::string& key)`
3. Update CMake to enable some aie2ps testing on Windows

Code cleanup and CMake updates

None

Build and ctest validated

NA

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,13 +86,13 @@ jobs:
     - name: Build (Linux)
       if: runner.os == 'Linux'
       run: |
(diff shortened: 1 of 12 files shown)


stsoe (changes requested):
LGTM

stsoe (approved):

PR #141, opened 2025-09-21
Framework for performing ctrlcode transforms

sonals:

1. New infrastructure which enables writing aie2p ctrlcode transforms using a pass manager.  p
2. Added a new utility, aiebu-transform to run the transforms

New feature

Added a new pass manager and a base class for defining a pass.  Currently the pass manager runs a canned sequence of transforms. However, in the final solution a JSON file will determine what transforms to run. 

**Note that the current ELF generated by aiebu cannot be read back and rewritten by aiebu. This issue needs to be solved first before the transforms can be enabled and saved. I have validated that with ELFIO I can read non aiebu (like x86 binaries) and then save it back. But reading and writing back ctrlcode ELF fails**
@HimanshuChoudhary-Xilinx and @donwalkarsoham please review the relevant code in transform.cpp to debug the underlying issue:
```
  ELFIO::elfio ebin;
  ebin.load(result["filename"].as<std::string>());
  // We fail in the save even without any transforms. First this needs to be
  // debugged and resolved before we can save any transformed ctrlcode. Once
  // this bug is fixed comment out the ebi.save() from here.
  ebin.save(result["output"].as<std::string>());
```

NA

Compile only tested


src/cpp/analyzer/passmanager.h
```diff
@@ -0,0 +1,53 @@
...
+  void adjust_relocations();
+
+public:
+  explicit passmanager(ELFIO::elfio &elf, bool debug) : m_elf(elf), m_debug(debug) {}
+
+  void run_transforms() {
+    // Currently we are running a precannded sequence of transforms.
+    ELFIO::Elf_Half sec_num = m_elf.sections.size();
+    for ( int i = 0; i < sec_num; ++i ) {
+      ELFIO::section *psec = m_elf.sections[i];
```
stsoe: Range based?

```suggestion
  for (auto psec : m_elf_sections) {
```

src/cpp/utils/transform/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+  ${AIEBU_SOURCE_DIR}/src/cpp/include
+  ${AIEBU_SOURCE_DIR}/src/cpp/cxxopts/include
+  ${AIEBU_SOURCE_DIR}/src/cpp
+  ${AIEBU_SOURCE_DIR}/src/cpp/ELFIO
+  ${AIEBU_SOURCE_DIR}/src/cpp/disassembler
+  ${AIEBU_SOURCE_DIR}/src/cpp/common
+  ${AIEBU_SOURCE_DIR}/src/cpp/ops
+  ${AIEBU_SOURCE_DIR}/src/cpp/preprocessor
+  ${AIEBU_SOURCE_DIR}/
+  ${Boost_INCLUDE_DIRS}
```
stsoe, line 20: PLease don't continue a bad trend.   Can we `#include` relative to say `${AIEBU_SOURCE_DIR}/src/cpp` ?   This makes it much easier for code readers to navigate to included file.

src/cpp/analyzer/basicpass.h
```diff
@@ -0,0 +1,60 @@
...
+};
+
+
+enum class basic_node_state {
+  original, // original ctrlcode node from input ELF is in pristine condition
+  dropped,  // original node from input ELF has been removed from the ctrlcode list
+  added     // a new node has been created and added to the ctrlcode list
+};
+
+template <typename aie2p_type> struct basic_node {
```
stsoe:
```suggestion
template <typename aie2p_type> 
struct basic_node {
```

src/cpp/utils/transform/transform.cpp
```diff
@@ -0,0 +1,99 @@
...
+  const std::vector<char> buffer = aiebu::readfile(result["filename"].as<std::string>());
+  if (aiebu::identify_buffer_type(buffer) !=
+      aiebu::aiebu_assembler::buffer_type::elf_aie2)
+    return 1;
+
+  ELFIO::elfio ebin;
+  ebin.load(result["filename"].as<std::string>());
+  // We fail in the save even without any transforms. First this needs to be
+  // debugged and resolved before we can save any transformed ctrlcode. Once
+  // this bug is fixed commen out the ebi.save() from here.
```
larry9523: @donwalkarsoham is trying to fix it by adding section sh_addr and segment p_vaddr by its size and alignment in ELF. Those apply to "PROGBITS" and "DYNAMIC" sections.

donwalkarsoham: Added PR https://github.com/Xilinx/aiebu/pull/149/files to fix issue from AIEBU side to avoid changes in ELFIO.
In aiebu we add virtual address of loadable segments so ELFIO::save() is able to layout the order of the segment sections correctly.

sonals:
Added a pre-transform operation which does the following:

1. Upgrades legacy ELF to add addresses to sections and segments -- the addresses match the physical offset
2. Remove ALLOC flag from all sections except text and data
3. Drops the  DYNAMIC segment. .dynamic section is enough
4. Changed the machine code to the official code, 269, allocated by ELF standards body

src/cpp/analyzer/passmanager.h
sonals: Done!

src/cpp/utils/transform/CMakeLists.txt
sonals: I have removed unnecessary paths.

src/cpp/analyzer/basicpass.h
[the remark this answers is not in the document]
sonals: Updated!

src/cpp/analyzer/basicpass.cpp
```diff
@@ -0,0 +1,392 @@
...
+  std::list<basic_node<XAie_OpHdr>> &m_nodes;
+  uint16_t m_pdiid;
+  uint64_t m_pdisize;
+
+public:
+  explicit XAie_OpHdr_add_loadpdi(std::list<basic_node<XAie_OpHdr>> &nodes, uint16_t pdiid, uint64_t pdisize)
+    : m_nodes(nodes), m_pdiid(pdiid), m_pdisize(pdisize) {}
+
+  void transform() override {
+    XAie_LoadPdiHdr *load = new XAie_LoadPdiHdr();
```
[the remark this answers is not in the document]
sonals: Done!

PR #143, opened 2025-09-26
Declare dependencies to fix parallel test runs

shramov:
Running tests in parallel on clean build always fail due to missing dependencies

How to reproduce:

```
rm -rf bld
mkdir bld
cd bld
cmake ..
make -j8
ctest --stop-on-failure --output-on-failure -j8
```

aie2ps_disassemble creates empty file when run before aie2ps_asm

Added missing dependencies


shramov:
@stsoe Can you take a look? This is needed to fix test failures when building XRT

sonals (approved):

Issue #144, opened 2025-09-27
Template used to generate isa.h header is outdated


shramov:
Latest update of specifications/aie2ps/isa.h intoroduced in #134 is not syncrhonized with templates/aie2ps/assembler_stub.h

When build system tries to regenerate header result is equivalent to commit before 3c4d2d8bf3c9e7720f415e1576ce18683f1dbe1b (normally isa-spec-gen target is not called but Ninja backend thinks that it is needed)

shramov:
@donwalkarsoham can you take a look?

lubosz:
I can reproduce this issue.
Every Ninja invocation after the first calls the `spec_tool.py` and generates a header with the `isa_disassembler` class missing, which is used in `disassembler.h`.

Weirdly it works in the first invocation of the build, where the header is not generated.

I can confirm that the `specification/aie2ps/isa.h` header is not regenerated at all when using the `-G "Unix Makefiles"` backend of cmake. Also for multiple invocations of the build.

This also breaks the XRT build for me, in case I invoke the build another time after the first.

It seems to be a build system bug that the header is only generated the second invocation of the build.

I suppose generation of this header should be either made mandatory or removed from the build system. In case the header needs to be regenerated on every system, the generation should not overwrite a header in the version controlled source directory but, it should be placed in the temporary build directory as build artifact.

lubosz:
This issue comes down to my initial suspicion of overwriting source files by build targets.

Ninja invokes the `cpp-assembler-stubs` phony target due to the definition of `BYPRODUCTS` in the `add_custom_target`.

This target is not automatically called for `all`, as intended, when no file from the source directory is in the BYPRODUCTS.

Commenting the `BYPRODUCTS` line avoids the issue.

According to the CMake documentation, `BYPRODUCTS` behave differently for Makefile and Ninja generators:
https://cmake.org/cmake/help/latest/command/add_custom_target.html

Another nasty outcome of defining a version controller source file as BYPRODUCTS is that it is deleted when calling the `clean` target with the Makefile generator 

```
cmake -G "Unix Makefiles" -B build-make
make -C build-make/ clean
```
This deletes a bunch of files in the `specification` directory including the header in question.

## Quick fix

Not specifying the version controlled header in question as byproduct avoids the issue by not executing the phony  `cpp-assembler-stubs`  target.

```diff
diff --git a/specification/aie2ps/CMakeLists.txt b/specification/aie2ps/CMakeLists.txt
[...]

PR #147, opened 2025-10-03
spec/build: Don't specify version controlled header as BYPRODUCTS.

lubosz:

This resolves an issue of the `isa.h` header being unintendedly regenerated when invoking the build a second time using the Ninja generator.

This resulted in a broken build due to isa_disassembler being undefined in the regenerated header.

Due to differences in CMake generators regarding to BYPRODUCS, the issue was not reproducible using the Unix Makefile generator

https://cmake.org/cmake/help/latest/command/add_custom_target.html

In general BYPRODUCTS is not intended to overwrite version controlled source files.

This patch addresses the symptom of issue #144.

Don't specify version controlled header as BYPRODUCTS.
A proper solution would be either not regenerating this header by the build system but by manually invoking the Python script and committing the result or not having the generated header in version control in case a regeneration is needed for every build. The current expected behavior did not require a regeneration for every build though.

More details can be found in #144.


lubosz:
This is still relevant even though #148 is merged.
The header is still being unnecessarily regenerated when building with Ninja.

PR #159, opened 2025-10-28
remove data section from dump

AdvaitNaik:
remove asm data section from elf dump section

NA

NA

NA

NA

NA

### DIFF
--- a/src/cpp/encoder/aie2ps/aie2ps_encoder.cpp
+++ b/src/cpp/encoder/aie2ps/aie2ps_encoder.cpp
@@ -194,9 +194,9 @@ page_writer(page& lpage, std::map<std::string, std::shared_ptr<scratchpad_info>>
       // TODO assert
     } else if (data->isOpcode())
     {
-      offset_type pc_low = pagenum * PAGE_SIZE + textwriter->tell() + datawriter->tell();
-      offset_type pc_high = pc_low + (*m_isa)[name]->serializer(data->get_operation()->get_args())->size(*page_state) - 1;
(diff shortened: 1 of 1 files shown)


src/cpp/encoder/aie2ps/aie2ps_encoder.cpp
```diff
@@ -194,9 +194,9 @@ page_writer(page& lpage, std::map<std::string, std::shared_ptr<scratchpad_info>>
       // TODO assert
     } else if (data->isOpcode())
     {
-      offset_type pc_low = pagenum * PAGE_SIZE + textwriter->tell() + datawriter->tell();
-      offset_type pc_high = pc_low + (*m_isa)[name]->serializer(data->get_operation()->get_args())->size(*page_state) - 1;
-      m_debug.add_dataline(fid, data->get_linenumber(), pc_high, pc_low, data->get_line(), data->get_annotation_index());
+      //offset_type pc_low = pagenum * PAGE_SIZE + textwriter->tell() + datawriter->tell();
+      //offset_type pc_high = pc_low + (*m_isa)[name]->serializer(data->get_operation()->get_args())->size(*page_state) - 1;
+      //m_debug.add_dataline(fid, data->get_linenumber(), pc_high, pc_low, data->get_line(), data->get_annotation_index());
```
donwalkarsoham, line 199: As per our discussion, if data section is needed in the future and we are disabling it temporarily, then we can add a flag to enable/disable the data section. This flag can also be made to pass during build time. This will avoid code rollback needed in future when the data section will be needed.

AdvaitNaik:
closing the PR, @HimanshuChoudhary-Xilinx will implement flag to enable dump section in elf at runtime and default can be disabled

PR #170, opened 2025-11-18
Performance optimization

HimanshuChoudhary-Xilinx:
Performance optimization
1. debug_map.json only generated in case of full dump flag.
2. Fast checks for regex
3. writer changes

NA

### DIFF
--- a/src/cpp/common/writer.cpp
+++ b/src/cpp/common/writer.cpp
@@ -27,6 +27,35 @@ write_word(uint32_t word)
   write_byte((word >> FORTH_BYTE_SHIFT) & BYTE_MASK);
 }
 
+// Bulk write methods
+void
+section_writer::
+write_bytes(const std::vector<uint8_t>& bytes)
+{
+  m_data.insert(m_data.end(), bytes.begin(), bytes.end());
+}
+
+void
+section_writer::
+write_bytes(const std::vector<char>& bytes)
+{
+  m_data.insert(m_data.end(), bytes.begin(), bytes.end());
+}
+
+void
+section_writer::
+write_bytes(const char* bytes, size_t count)
+{
+  m_data.insert(m_data.end(), bytes, bytes + count);
+}
+
+void
+section_writer::
+reserve(size_t capacity)
+{
+  m_data.reserve(capacity);
+}
+
 offset_type
 section_writer::
 tell() const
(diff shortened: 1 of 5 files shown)


src/cpp/common/writer.cpp
```diff
@@ -27,6 +27,35 @@ write_word(uint32_t word)
...
+  m_data.insert(m_data.end(), bytes.begin(), bytes.end());
+}
+
+void
+section_writer::
+write_bytes(const char* bytes, size_t count)
+{
+  m_data.insert(m_data.end(), bytes, bytes + count);
+}
+
```
sonals, line 30: Can we templatize write_bytes() function? 

```
template <typename T> void
section_writer::
write_bytes(const T &bytes)
{
  m_data.insert(m_data.end(), bytes.begin(), bytes.end());
}
```
Also you can get rid of write_bytes(const char* bytes, size_t count) which is called from only place:
`dumpwriter->write_bytes(dbg_str.c_str(), dbg_str.size());`
With the template method we can use the dbg_str directly.

src/cpp/preprocessor/asm/asm_parser.cpp
```diff
@@ -163,8 +162,8 @@ parse_lines(const std::vector<char>& data, std::string& file)
       continue;
     }
 
-    // check for label
-    if (boost::regex_match(line, sm, LABEL_REGEX))
+    // check for label - fast check for ':' at end before regex
+    if (line.back() == ':' && boost::regex_match(line, sm, LABEL_REGEX))
     {
```
sonals, line 167: Can we quantify how much improvement we get with the micro-optimizations above?

HimanshuChoudhary-Xilinx: Release build with line 139 and 166 get ~2sec improvement of DMA compiler test

src/cpp/common/writer.cpp
HimanshuChoudhary-Xilinx: done

src/cpp/preprocessor/asm/asm_parser.cpp
sonals: Wow!

sonals (approved):

PR #175, opened 2025-11-21
Add min supported version for Ubuntu and Boost

donwalkarsoham:
Post std::regex to boost::regex migration, Boost 1.76 or above is needed as per below documentation. https://www.boost.org/releases/1.76.0/

N/A

Adding minimum boost and Ubuntu version needed in Readme.rst

N/A

Used readme_renderer to convert to html and validate the changes.

Readme.rst

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,7 +22,7 @@ Build Dependencies
 
  * cmake 3.18 or above
  * c++17 compiler
- * Boost (header only)
+ * Boost (header only) minimum version supported is 1.76
  * cxxopts (included as submodule)
  * ELFIO (included as submodule)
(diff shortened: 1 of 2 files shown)


sonals (changes requested):
Can you add a boost version check in cmake/boost.cmake where we call `find_package(Boost...`

donwalkarsoham:
> Can you add a boost version check in cmake/boost.cmake where we call `find_package(Boost...`

Currently, I have enforced requirement of using Boost 1.76 or above in cmake/boost.cmake by letting the build fail if otherwise. Let me know if this is fine.

sonals (approved):

PR #179, opened 2025-11-26
Add regex wrapper

donwalkarsoham:
Some customers are still using Ubuntu 22.04 which has older boost versions (<1.76 ) and don't support Boost::regex header only.
Hence, for customers having boost < 1.76 will be using std::regex and boost>=1.76 will use boost::regex

N/A

1> Added regex wrapper in aiebu wrapped around std::regex and boost::regex. boost::regex v/s std::regex  is selected considering boost::regex version.
```
If boost version < 1.76 
    std::regex
Else
   boost::regex
```

N/A

1> Successfully built and ran build ctests on Windows and Ubuntu 22.04
2> Ubuntu CI pipelines build ran successfully too which is Ubuntu 24.04

### DIFF
--- a/src/cpp/common/regex_wrapper.h
+++ b/src/cpp/common/regex_wrapper.h
@@ -1,0 +1,66 @@
+#ifndef AIEBU_COMMON_REGEX_WRAPPER_H_
+#define AIEBU_COMMON_REGEX_WRAPPER_H_
+
+/**
+ * @file regex_wrapper.h
+ * @brief Unified regex API wrapper supporting both std::regex and boost::regex
+ * 
+ * This wrapper allows transparent switching between std::regex and boost::regex
+ * based on platform requirements:
+ * - Ubuntu 22.04 and earlier: uses std::regex (better integration with system)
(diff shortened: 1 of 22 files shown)


src/cpp/common/regex_wrapper.h
```diff
@@ -0,0 +1,66 @@
...
+ * - Windows: uses boost::regex (consistent with Boost ecosystem)
+ * 
+ * The wrapper is controlled by the USE_BOOST_REGEX CMake option.
+ */
+
+#ifdef USE_BOOST_REGEX
+  #include <boost/regex.hpp>
+#else
+  #include <regex>
+#endif
```
stsoe, line 21: nit:
```suggestion
#ifdef USE_BOOST_REGEX
# include <boost/regex.hpp>
#else
# include <regex>
#endif
```

stsoe (approved):

HimanshuChoudhary-Xilinx (approved):

sonals (approved):

PR #181, opened 2025-12-01
dynamic tracing compiler warnings clean up

AdvaitNaik:

dynamic tracing compiler warnings clean up


src/cpp/dtrace/dtrace.cpp
```diff
@@ -249,10 +218,10 @@ get_dtrace_result_file(const char* result_file)
             // Get control_buffer and mem_buffer from the buffer
             std::vector<uint32_t> control_buffer(
                 buffer.begin(),
-                buffer.begin() + l_dtrace_buffer_info.control_buffer.size()
+                buffer.begin() + l_dtrace_buffer_info.control_buffer.size() // NOLINT
```
sonals: Please use the full syntax to document what we are suppressing.
`// NOLINT ( check-name-list ) `
Grep for NOLINT in src/cpp to see examples

sonals (changes requested):
Please document the rule in `// NOLINT(rule)`  for the non obvious cases.

src/cpp/dtrace/dtrace.cpp
AdvaitNaik: Done

PR #184, opened 2025-12-02
Added kernel:instance filter for get tlb api

HimanshuChoudhary-Xilinx:
Update argtbl get_argtbl() api to take kernel:instance name for full elf.
It return tbl only for that kernel instance.
If we use xclbin + elf flow ... elf will not have kernel info so call get_argtlb() without kernel name.

### DIFF
--- a/src/cpp/include/aiebu/aiebu_assembler.h
+++ b/src/cpp/include/aiebu/aiebu_assembler.h
@@ -181,14 +181,24 @@ class aiebu_assembler
          * xrt_idx transform.
          */
         std::vector<arginfo>& dump() const;
+
+        /*
+         * Get the kernel:instance name filter used for this argtbl
+         * @return const reference to the name string (e.g., "DPU:dpu")
+         */
         const std::string& get_name() const;
     };
+
     /*
-     * Get an argtbl object from aiebu_assembler. In this function
-     * aiebu_assember will scan the ELF and construct a table (vector)
-     * of xrt_idx and BD offset. Then it returns an object of this table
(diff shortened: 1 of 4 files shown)


src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -190,7 +190,7 @@ class aiebu_assembler
      * which can be used to dump the reference of the table and modify
      * the table in place.
      */
-    argtbl get_argtbl();
+    argtbl get_argtbl(const std::string& name = "");
```
larry9523, line 203: Please update the comment (doc) of this API accordingly. I've updated the same on the spec page.

src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -181,7 +181,7 @@ class aiebu_assembler
          * xrt_idx transform.
          */
         std::vector<arginfo>& dump() const;
-
+        const std::string& get_name() const;
```
larry9523, line 189: We need to put some comment on this public API

HimanshuChoudhary-Xilinx: added

src/cpp/include/aiebu/aiebu_assembler.h
HimanshuChoudhary-Xilinx: added

HimanshuChoudhary-Xilinx:
Closing as we raised https://github.com/Xilinx/aiebu/pull/186 with new spec

PR #186, opened 2025-12-03
Kernel name change api

HimanshuChoudhary-Xilinx:
1. Update argtbl get_argtbl() api to take kernel name for full elf.
2. It return tbl only for all instances in that kernel.
3. These apis applicable only for aie2ps/aie4 config/full elf's
4. Added testcase
5. NOTE: if kernel have 2 instance like in example, kernel name for both instance modify if we modify for one instance like in example.

Usage:
```
auto as = aiebu::aiebu_assembler(elf_buf);
auto tbl = as.get_argtbl("DPU");
auto& inst = tbl.get();                     // this should be ref so changes are reflected
auto& table = inst[0].inst_arginfo;  // this should be ref so changes are reflected
tbl.set_name("RESNET");
table[1].xrt_idx = 45;
table[1].bd_offset = 100;
....
as.flush_argtbl(tbl);
auto e_modified = as.get_elf();
```
```
kernel info in ORIGINAL ELF
Symbol table '.symtab' contains 3 entries:
   Num:    Value  Size Type    Bind   Vis      Ndx Name
     0: 00000000     0 NOTYPE  LOCAL  DEFAULT  UND 
     1: 00000000     0 FUNC    WEAK   DEFAULT  UND _Z3DPUPcPcPcPc
     2: 00000000     0 OBJECT  WEAK   DEFAULT    1 vad123
     3: 00000000     0 OBJECT  WEAK   DEFAULT    1 vadd
```
```
kernel info in MODIFIED ELF

Symbol table '.symtab' contains 3 entries:
   Num:    Value  Size Type    Bind   Vis      Ndx Name
     0: 00000000     0 NOTYPE  LOCAL  DEFAULT  UND 
     1: 00000000     0 FUNC    WEAK   DEFAULT  UND _Z6RESNETPcPcPcPc
     2: 00000000     0 OBJECT  WEAK   DEFAULT    1 vad123
     3: 00000000     0 OBJECT  WEAK   DEFAULT    1 vadd

```


src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -152,53 +164,81 @@ class aiebu_assembler
...
+      * Get an argtbl object from aiebu_assembler for a given
+      * kernel name. In this function aiebu_assember will scan
+      * the ELF and construct a vector of instance info (instinfo).
+      * Inside each instinfo, we have a table (vector)
+      * of xrt_idx and BD offset of that kernel:inst's control code.
+      *
+      * The object can be used to dump the reference of the instance
+      * info and modify the table in place inside each instance.
+      *
+      * NOTE: applicable for only config elf's
```
larry9523: We don't really have a official concept of "config elf's". Are you trying to say it is only available for full ELF flow?

src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -152,53 +164,81 @@ class aiebu_assembler
...
-    /*
-     * Flush the argtbl object to aiebu_assembler. In this function,
-     * aiebu will take the argtbl object and update control code and
-     * patching metadata based on the xrt_idx and BD offset in the
-     * arg table.
-     */
-    void flush_argtbl(const argtbl& arg_table);
+     class argtbl
+     {
+         friend class aiebu_assembler;
```
larry9523: why do we need this?

HimanshuChoudhary-Xilinx: i want to use kernel name handle in aiebu_assembler class to get old and new kernel name

src/cpp/include/aiebu/aiebu_assembler.h
HimanshuChoudhary-Xilinx: yes ... updated

src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -152,53 +164,81 @@ class aiebu_assembler
...
+      * the ELF and construct a vector of instance info (instinfo).
+      * Inside each instinfo, we have a table (vector)
+      * of xrt_idx and BD offset of that kernel:inst's control code.
+      *
+      * The object can be used to dump the reference of the instance
+      * info and modify the table in place inside each instance.
+      *
+      * NOTE: applicable for only full elf's
+      */
+     argtbl get_argtbl(const std::string kernel_name);
```
stsoe:
```suggestion
     argtbl get_argtbl(const std::string& kernel_name);
```

src/cpp/include/aiebu/aiebu_assembler.h
stsoe: Handle classes can safely return the handle through a public inline API (get_handle()).  The handle is useless to public clients, but Implementation code can then access the impl this way.   It accomplishes the same as the friend class though, so really no preference one way or the other.

FWIW, XRT uses get_handle for all pimpls, see here https://github.com/Xilinx/XRT/blob/5c6def2df468485b7240e3749d94994f0deeeb2f/src/runtime_src/core/include/xrt/detail/pimpl.h#L52

larry9523 (approved):

sonals:
Can you update the PR description to say 
`tbl.set_name("RESNET"); ` instead of 
`tbl.set_name("CONV");` as the readelf output for modified elf is printing` _Z6RESNETPcPcPcPc`

src/cpp/analyzer/transform_manager.h
```diff
@@ -211,27 +245,68 @@ class transform_manager {
...
+   * @return Modified ELF binary as vector of chars
+   *
+   * For C++ mangled names, matches the exact identifier (e.g., "DPU" matches "_Z3DPUPcPc"
+   * but not "_Z4DPU1PcPc"). Automatically updates length prefixes (_Z3DPU -> _Z4XCVB).
+   * For non-mangled names, does exact string matching.
+   * Only supports OS ABI 0x46 and ABI version 0x3.
+   *
+   * @throws error if format unsupported, sections missing, or name not found
+   */
+   std::vector<char> update_kernel_name(const std::string orig_name, const std::string new_name);
```
sonals: std::string reference?

src/cpp/assembler/aiebu_assembler.cpp
```diff
@@ -122,52 +122,107 @@ disassemble(const std::filesystem::path &root) const
...
+set_name(const std::string& name)
+{
+  handle->set_name(name);
 }
 
 aiebu_assembler::argtbl
 aiebu_assembler::
-get_argtbl()
+get_argtbl(const std::string kernel_name)
 {
```
sonals, line 188: reference?

sonals (changes requested):
Please update the code to use std::string references where applicable.

src/cpp/include/aiebu/aiebu_assembler.h
HimanshuChoudhary-Xilinx: done

PR #189, opened 2025-12-09
In mem assembling

sayyanna:
Add dedicated full elf APIs for aie2, aie2ps/aie4
https://amd.atlassian.net/wiki/spaces/AIE/pages/1178885648/AIEBU+Full+elf+apis
Added dedicated full elf APIs for aie2, aie2ps/aie4 for in-memory assembling
None
Added testcases, and compared the checksum with file system based elf files that are generated using command line. Verification on target is in progress.


src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -26,6 +27,32 @@ struct arginfo {
   uint64_t bd_offset;
 };
 
+class file_artifact_impl;
+class file_artifact
```
stsoe, line 47: I looked at this class for some time before I realized its purpose.  This being a public header file, please make sure to document the purpose and how to to use it in the header file.  How are any users going to figure out what this is?

src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -26,6 +27,32 @@ struct arginfo {
...
+{
+  public:
+    file_artifact();
+    ~file_artifact();
+    /**
+     * Add a virtual file into the artifact by reference.
```
stsoe: What is a "virtual file"?

src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -26,6 +27,32 @@ struct arginfo {
...
+     * owns the name and buffer
+     */
+    void add_vfile(const std::string& name, const std::vector<char>& buffer);
+    /**
+     * Add a virtual file into the artifact by rvalue.
+     *
+     * Note: there are no extra copy of name and the buffer. But the ownership
+     * of them are transferred.
+     */
+    void add_vfile(std::string&& name, std::vector<char>&& buffer);
```
stsoe: rvalue of buffer makes sense, I am not so sure about the name?

src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -26,6 +27,32 @@ struct arginfo {
...
+
+    const std::vector<char>& get(const std::string& filename) const;
```
stsoe: Public function in exported API must be documented.  What is "filename", is that the same as the "name" used in the add functions?   It would be good to know that this class represents in-memory data that is identified by a "key", if that is really what it is?  Is this class specific to handling file data?  I don't think so, so why is it called "file_artifact"?  

If the class is specific to files, or if the most common use case is files,  then why does it not have a function to read a file from disk and cache it's content?    Do the clients really have to read the file first?  It's fine that it offers the capability of taking the user's data, but maybe in some cases it should just get the data directly from a file?

src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -114,6 +141,21 @@ class aiebu_assembler
...
+     * 
+     * @type:               ELF buffer type (aie2_config, aie2ps_config, aie4_config)
+     * @config_json_buffer: Config json content
+     * @artifact:           file_artifact object contains the mapping between file path
+     *                      and its binary
+     */     
+    aiebu_assembler(buffer_type type,
+                    const std::vector<char>& config_json_buffer,
+                    file_artifact& artifact,
+                    const std::vector<std::string>& flags);
```
stsoe, line 197: I am looking at this constructor in isolation and I have no idea what "flags" is?

src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -26,6 +27,49 @@ struct arginfo {
...
+    /*
+     * Add a virtual file (in-memory buffer) into the artifact by rvalue.
+     *
+     * Note: there is no extra copy of the buffer. But the ownership
+     * of the buffer is transferred.
+     *
+     * @param name   name of the buffer or virtual file.
+     * @param buffer contents stored as a vector of chars.
+     */
+    void add_vfile(std::string& name, std::vector<char>&& buffer);
```
stsoe, line 70:
```suggestion
    void add_vfile(const std::string& name, std::vector<char>&& buffer);
```

src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -114,6 +158,24 @@ class aiebu_assembler
...
+     * @artifact:           file_artifact object contains the mapping between
+     *                      virtual file (in-memory buffer) name and its binary
+     * @flags:              for passing configuration flags to the assembler
+     *                      ex: disabledump (disable debug dump),
+     *                          fulldump (enable debug dump),
+     *                          opt_level_1 (for optimization)
+     */
+    aiebu_assembler(buffer_type type,
+                    const std::vector<char>& config_json_buffer,
+                    file_artifact& artifact,
```
stsoe:
```suggestion
                    const file_artifact& artifact,
```

stsoe (commented):
The code could be simpler.  There are many instances with `if (artifacts) ... else ...` .  You should change the implementation of the public API in aiebu_assembler to always create a file_artifacts before it calls into the real implementation code.

If the user does not provide a file_artifacts, then you create one.  You amend file_artifacts with a function that take the name of a file.  If the file is present in the artifacts then return its content, otherwise read the file and move it into the artifacts, then return the content.   This way you always have a valid `file_artifacts` in all functions and you avoid `if ... else ...` 

The implementation of the file_artifacts uses a map to store the key, value.  This is good.  The implementation is logically constant, meaning that any artifacts added by a user who creates a `file_artifacts` remain in the map.   But, if you follow my suggestion above, your AIEBU implementation can add file to the artifacts as it moves a long.   In order to allow the user level API to take `const file_artifacts&`  you should make the underlying map `mutable` so that your implementation code can add files to the repo via a const method. 

You may want to add the artifacts passed by the user as a `const file_artifacts* m_artifacts` in all classes that need it so as not having to always pass it as a function argument.

One more note, if we are worried about memory consumption of transient one-time read files and don't want to store them in the artifacts, e.g. read the file, skip the move into the artifacts map, but return its content, then the reference to std::vector presents a problem.    You probably can't solve this without some if/else condition and a separate API.  I am not sure this is a valid concern given that the the application can anyway pass all the files in-memory or none at all.  Do we have the none-at-all path for a memory consumption reason?

sonals:
Please rebase and make the changes suggested. Do not check-in binary files into the repository. For .bin files see the approach used in the tests -- the files are base64 encoded and then the ASCII file is check in.

sonals:
Please post an updated PR with conflicts resolved.

sayyanna:
> The code could be simpler. There are many instances with `if (artifacts) ... else ...` . You should change the implementation of the public API in aiebu_assembler to always create a file_artifacts before it calls into the real implementation code.
> 
> If the user does not provide a file_artifacts, then you create one. You amend file_artifacts with a function that take the name of a file. If the file is present in the artifacts then return its content, otherwise read the file and move it into the artifacts, then return the content. This way you always have a valid `file_artifacts` in all functions and you avoid `if ... else ...`
> 
> The implementation of the file_artifacts uses a map to store the key, value. This is good. The implementation is logically constant, meaning that any artifacts added by a user who creates a `file_artifacts` remain in the map. But, if you follow my suggestion above, your AIEBU implementation can add file to the artifacts as it moves a long. In order to allow the user level API to take `const file_artifacts&` you should make the underlying map `mutable` so that your implementation code can add files to the repo via a const method.
> 
> You may want to add the artifacts passed by the user as a `const file_artifacts* m_artifacts` in all classes that need it so as not having to always pass it as a function argument.
> 
> One more note, if we are worried about memory consumption of transient one-time read files and don't want to store them in the artifacts, e.g. read the file, skip the move into the artifacts map, but return its content, then the reference to std::vector presents a problem. You probably can't solve this without some if/else condition and a separate API. I am not sure this is a valid concern given that the the application can anyway pass all the files in-memory or none at all. Do we have the none-at-all path for a memory consumption reason?

Do we really need to move the file contents into `file_artifacts`? My understanding is that once the content is read, we don’t read the same file again. If caching the file content is actually required, I can modify the implementation accordingly.

sayyanna:
> Please rebase and make the changes suggested. Do not check-in binary files into the repository. For .bin files see the approach used in the tests -- the files are base64 encoded and then the ASCII file is check in.

Getting some error with this command: cmake -P b64.cmake -e <input.bin> , will upload the encoded files shortly
cmake -P b64.cmake -e ml_txn_lp.bin ml_txn_lp.b64
CMake Error: Error processing file: b64.cmake

src/cpp/include/aiebu/aiebu_assembler.h
```diff
@@ -72,7 +124,7 @@ class aiebu_assembler
     buffer_type m_type;
     buffer_type m_output_type;
     class argtbl_impl;  // Forward declaration
-
+    file_artifact artifacts;
```
stsoe, line 127: This is the problem with not insulating public API classes.  We are now adding a new data member and have set ourself up for breaking backwards compatibility (ABI).  What if any existing binary is compiled against the version before artifacts was added?   It will not work because the artifacts member as seen by implementation, will be bogus (it will not exist) in the binary.  We cannot do this.   

You cannot change the layout of public facing classes.   I hate making this difficult, someone need to fix this,  but not part of your changes.   

What you need to do is making sure not to add any new data members to public facing classes.   You will have to carry the artifacts argument from the new public API through to the implementation of the new API(s).  In the implementation of the new public API, you can pass the artifacts to internal classes and set as data member there if necessary to simply further passing of the artifacts pointer.

stsoe (changes requested):
See comment on new data member.  This will not work as is.
Actually, I take this back.   Since AIEBU is statically linked with all clients, there is never a chance for ABI breakage.

src/cpp/include/aiebu/aiebu_assembler.h
stsoe: @sayyanna You can ignore the above comment about adding a new data member.  It will work here in this case because all clients are statically linking with AIEBU.  That doesn't mean the design of the public facing classes is good, but at least it is safe.

larry9523: @sayyanna in the spec page https://amd.atlassian.net/wiki/spaces/AIE/pages/1178885648/AIEBU+Full+elf+apis
we only add a new method in aiebu_assembler class.

I also have a simple prototype to prove this idea
/proj/xsjhdstaff5/larryliu/testcode/cpp_rampath

stsoe (approved):
Looks fair now given that clients are statically linking with AIEBU.
Internal passing of `const file_artifacts*` is okay even as other arguments are passed by reference.  The reference arguments are passed from clients but file_artifacts is managed by the class public facing class, so this distinction is good.

sonals (changes requested):
Please remove the binary PDI files from the PR.

sonals (approved):

PR #196, opened 2025-12-13
Add binary i/p support to disassembler + fix bugs

donwalkarsoham:
1> Add support to disassemble a binary input to ASM
2> Fix sign-extension bug leading to value overflow
3> .align frequency
4> Some small values were displayed in hex (0x..)which should decimal.

1> Added binary parser
2> type casted to unsigned integer as current type was char hence was signed leading to overflow.
Old value - 0xFFFFFFCD
Expected/corrected value - 0xabcdabcd
3> .align was repeated after every label. Fixed it to be only once before the start of sections.
4> Added logic to display small values in decimal if bigger values are present then display in hex. This is as per current ASM
format seen.

N/A

1> Dumped single_col_vadd binaries in ASM format.
2> Dumped single_col_vadd and sinle_col_move_memtiles ELF in ASM format

### DIFF
--- a/src/cpp/analyzer/reporter.cpp
+++ b/src/cpp/analyzer/reporter.cpp
@@ -182,6 +182,16 @@ namespace aiebu {
         else if (m_buffer_type == aiebu::aiebu_assembler::buffer_type::blob_instr_transaction) {
             disassemble_blob(stream);
         }
+        else if (m_buffer_type == aiebu::aiebu_assembler::buffer_type::unspecified) {
+            // Treat unspecified buffer type as raw binary data
+            try {
+                aiebu::asm_disassembler disasm(m_buffer, stream, true);
+                disasm.run();
+            } catch (const std::exception& ex) {
+                throw error(error::error_code::internal_error,
+                    "Binary disassembler error: " + std::string(ex.what()));
+            }
+        }
         else {
             throw error(error::error_code::invalid_buffer_type,
                   "Invalid buffer type for disassembly");
--- a/src/cpp/analyzer/reporter.cpp
+++ b/src/cpp/analyzer/reporter.cpp
@@ -14,8 +14,8 @@
 #include <elfio/elfio_dump.hpp>
 namespace aiebu {
 
-    reporter::reporter(aiebu::aiebu_assembler::buffer_type type, const std::vector<char>& buffer, const std::string& target_arch) :
-        m_buffer_type(type), m_buffer(buffer), m_target_arch(target_arch)
+    reporter::reporter(aiebu::aiebu_assembler::buffer_type type, const std::vector<char>& buffer) :
+        m_buffer_type(type), m_buffer(buffer)
     {
         if (buffer.empty()) {
             throw error(error::error_code::invalid_input, "Input buffer is empty");
@@ -132,7 +132,7 @@ namespace aiebu {
         }
         else if (m_buffer_type == aiebu::aiebu_assembler::buffer_type::elf_aie2ps ) {
             try {
-                aiebu::asm_disassembler disasm(root.string(), std::cout);
+                aiebu::elf_asm_disassembler disasm(root.string(), std::cout);
                 disasm.run();
             } catch (const std::exception& ex) {
                 throw error(error::error_code::internal_error,
@@ -179,10 +179,8 @@ namespace aiebu {
                 stream << tprint.get_all_ops() << std::endl;
             }
         }
-        else if (m_buffer_type == aiebu::aiebu_assembler::buffer_type::blob_instr_transaction) {
-            disassemble_blob(stream);
-        }
-        else if (m_buffer_type == aiebu::aiebu_assembler::buffer_type::blob_aie2ps ||
+        else if (m_buffer_type == aiebu::aiebu_assembler::buffer_type::blob_instr_transaction ||
+                 m_buffer_type == aiebu::aiebu_assembler::buffer_type::blob_aie2ps ||
                  m_buffer_type == aiebu::aiebu_assembler::buffer_type::blob_aie4) {
             // Disassemble binary files - delegate to disassemble_blob for evaluation
             disassemble_blob(stream);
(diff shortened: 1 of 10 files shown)


src/cpp/disassembler/disassembler.cpp
```diff
@@ -13,15 +15,27 @@ static constexpr size_t CTRLTEXT_STRING_LENGTH = 9;  // Length of ".ctrltext"
...
-    : m_asm_writer(output_stream) {
+    : m_asm_writer(output_stream), m_is_binary_mode(false) {
     if (!m_elf_reader.load(input_elf_path)) {
         throw error(error::error_code::invalid_elf, "Failed to load ELF:" + input_elf_path + "\n");
     }
     isa_op_map = isa_disasm.get_isa_map();
 }
 
+asm_disassembler::asm_disassembler(const std::vector<char>& binary_data, std::ostream& output_stream, bool is_binary)
+    : m_asm_writer(output_stream), m_binary_data(binary_data), m_is_binary_mode(is_binary) {
```
sonals: Why do we need `is_binary` parameter?

src/cpp/disassembler/disassembler.cpp
```diff
@@ -51,7 +65,7 @@ void asm_disassembler::print_section_info(const ELFIO::section* section) {
         flags += "x";
     m_asm_writer.write_directive("");
     if (is_data_section(section->get_name()))
-        m_asm_writer.write_directive(".align " + std::to_string(section->get_addr_align()));
+        m_asm_writer.write_directive("  .ALIGN             " + std::to_string(section->get_addr_align()));
```
sonals: Why upper case?

src/cpp/disassembler/disassembler.cpp
```diff
@@ -86,7 +101,11 @@ void asm_disassembler::process_data_section(const ELFIO::section* section, std::
             size_t consumed = deserializer.deserialize(m_asm_writer, state, section_data + offset);
             offset += consumed;
         } else if (local_ptr_map.find(state->get_address()) != local_ptr_map.end()) {
-            m_asm_writer.write_directive(".align 4");
+            if (!align_4_written) {
+                m_asm_writer.write_directive("");
+                m_asm_writer.write_directive("  .ALIGN             4");
```
sonals: ditto

src/cpp/disassembler/disassembler.cpp
donwalkarsoham: Done

src/cpp/disassembler/disassembler.cpp
donwalkarsoham: As per compiler generated ASM format.

src/cpp/disassembler/disassembler.cpp
donwalkarsoham: As per compiler generated ASM format.

src/cpp/analyzer/reporter.cpp
```diff
@@ -14,8 +14,8 @@
 #include <elfio/elfio_dump.hpp>
 namespace aiebu {
 
-    reporter::reporter(aiebu::aiebu_assembler::buffer_type type, const std::vector<char>& buffer) :
-        m_buffer_type(type), m_buffer(buffer)
+    reporter::reporter(aiebu::aiebu_assembler::buffer_type type, const std::vector<char>& buffer, const std::string& target_arch) :
+        m_buffer_type(type), m_buffer(buffer), m_target_arch(target_arch)
```
sonals: What is the purpose of `target_arch`? Shouldn't `buffer_type` be sufficient?

src/cpp/disassembler/disassembler.cpp
sonals: We follow GNU as syntax https://sourceware.org/binutils/docs/as.html#Align. The test examples also use lower case. File a CR against the compiler which is generating directive in upper case.

src/cpp/analyzer/reporter.cpp
```diff
@@ -182,6 +182,11 @@ namespace aiebu {
         else if (m_buffer_type == aiebu::aiebu_assembler::buffer_type::blob_instr_transaction) {
             disassemble_blob(stream);
         }
+        else if (m_buffer_type == aiebu::aiebu_assembler::buffer_type::blob_aie2ps ||
+                 m_buffer_type == aiebu::aiebu_assembler::buffer_type::blob_aie4) {
+            // Disassemble binary files - delegate to disassemble_blob for evaluation
+            disassemble_blob(stream);
+        }
```
sonals, line 187: Please merge this _else_ with the previous `else-if` since both of them have the same body.

src/cpp/analyzer/reporter.cpp
donwalkarsoham: Ohh yes, looks like older version of the code got submitted while switching between Windows and Linux testing. Let me correct and submit.

src/cpp/disassembler/disassembler.cpp
```diff
@@ -4,24 +4,73 @@
...
+static constexpr size_t ALIGN_16 = 16;                      // 16-byte alignment
+
+// Section Name Lengths
+static constexpr size_t CTRLTEXT_STRING_LENGTH = 9;        // Length of ".ctrltext"
+static constexpr size_t CTRLDATA_STRING_LENGTH = 9;        // Length of ".ctrldata"
 
 asm_disassembler::asm_disassembler(const std::string& input_elf_path, std::ostream& output_stream)
-    : m_asm_writer(output_stream) {
+    : m_asm_writer(output_stream), m_is_binary_mode(false), m_target_arch("aie2ps") {
     if (!m_elf_reader.load(input_elf_path)) {
```
sonals, line 141: What are `m_is_binary_mode`  and `m_target_arch` needed?

src/cpp/disassembler/disassembler.cpp
```diff
@@ -85,12 +135,16 @@ void asm_disassembler::process_data_section(const ELFIO::section* section, std::
             ucDmaBd_op_deserializer deserializer(&dummy_isa_op);
             size_t consumed = deserializer.deserialize(m_asm_writer, state, section_data + offset);
             offset += consumed;
-        } else if (local_ptr_map.find(state->get_address()) != local_ptr_map.end()) {
-            m_asm_writer.write_directive(".align 4");
+        }         else if (local_ptr_map.find(state->get_address()) != local_ptr_map.end()) {
```
sonals: Indentation

src/cpp/disassembler/disassembler.cpp
donwalkarsoham: This is not needed, part of an older code version where I didn't use buffer_type for arch. I will correct and submit latest changes.

src/cpp/analyzer/reporter.cpp
donwalkarsoham: done

src/cpp/disassembler/disassembler.cpp
donwalkarsoham: done

src/cpp/disassembler/disassembler.cpp
donwalkarsoham: Added newer version removing this plus added inheritance.

sonals (approved):

PR #204, opened 2026-01-06
update uuid on updating controlcode

HimanshuChoudhary-Xilinx:
1. update uuid on updating controlcode
2. bugfix: If offset in apply_offset_57 when xrt_id is greater than 6


larry9523 (commented):
The uuid part looks good.

Regarding xrt_id > 6, two questions

Does the limitation come from # of global register?
what is the protocol here that we leave the xrt_id (>6) in APPLY_OFFSET_57? Will CERT fail gracefully if it sees the number?

HimanshuChoudhary-Xilinx:
> The uuid part looks good.
> 
> Regarding xrt_id > 6, two questions
> 
> Does the limitation come from # of global register? what is the protocol here that we leave the xrt_id (>6) in APPLY_OFFSET_57? Will CERT fail gracefully if it sees the number?

Yes, the limitation come from global register.
for any xrt_id > 6, cert will cert skip patching.

larry9523:
> > The uuid part looks good.
> > Regarding xrt_id > 6, two questions
> > Does the limitation come from # of global register? what is the protocol here that we leave the xrt_id (>6) in APPLY_OFFSET_57? Will CERT fail gracefully if it sees the number?
> 
> Yes, the limitation come from global register. for any xrt_id > 6, cert will cert skip patching.

Then we should use a macro instead of magic number (or at least comment it).

And, it is not very clear to me who and how to use this code->offset. 
if xrt_id = 4, code->offset is 8
if xrt_id = 8, code->offset is 8
how to differentiate between them?

HimanshuChoudhary-Xilinx:
> > > The uuid part looks good.
> > > Regarding xrt_id > 6, two questions
> > > Does the limitation come from # of global register? what is the protocol here that we leave the xrt_id (>6) in APPLY_OFFSET_57? Will CERT fail gracefully if it sees the number?
> > 
> > 
> > Yes, the limitation come from global register. for any xrt_id > 6, cert will cert skip patching.
> 
> Then we should use a macro instead of magic number (or at least comment it).
> 
> And, it is not very clear to me who and how to use this code->offset. if xrt_id = 4, code->offset is 8 if xrt_id = 8, code->offset is 8 how to differentiate between them?

i get you point ... i think we should keep it uniform and fix in encoding. We should always send offset (= 2*xrt_id).

larry9523 (approved):

PR #207, opened 2026-01-08
added dump for 0.1 version of txnbin

HimanshuChoudhary-Xilinx:
```aiebu-dump -d -m aie2txn  <ml_txn.bin>```
was dumping only optimized (1.0version) txnbin 
Added fix for 0.1 version of txnbin

### DIFF
--- a/src/cpp/common/file_utils.cpp
+++ b/src/cpp/common/file_utils.cpp
@@ -13,7 +13,10 @@ constexpr unsigned int magic_length = 16;
 constexpr unsigned int elf_magic = 0x464c457f;
 
 // TODO: Add magic numbers for other AIE flavors
-constexpr unsigned int ctrlcode_magic_aie2 = 0x06040001;
+// optimized header major:1 minor:0 col:4 row:6
+constexpr unsigned int ctrlcode_magic_aie2_opt = 0x06040001;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/cpp/common/file_utils.cpp [hunk 1]
```diff
@@ -13,7 +13,10 @@ constexpr unsigned int magic_length = 16;
 constexpr unsigned int elf_magic = 0x464c457f;
 
 // TODO: Add magic numbers for other AIE flavors
-constexpr unsigned int ctrlcode_magic_aie2 = 0x06040001;
+// optimized header major:1 minor:0 col:4 row:6
+constexpr unsigned int ctrlcode_magic_aie2_opt = 0x06040001;
+// non-optimized header major:0 minor:1 col:4 row:6
+constexpr unsigned int ctrlcode_magic_aie2 = 0x06040100;
```
larry9523: I think the "04" in the so called magic number is not column. It is the generation. #col varies between control code.

larry9523 (changes requested):
Please fix the comment.

src/cpp/common/file_utils.cpp
HimanshuChoudhary-Xilinx: fixed

larry9523 (approved):

PR #212, opened 2026-01-15
dynamic tracing compiler action enhancement 

AdvaitNaik:

dynamic tracing compiler action enhancement 
- host timestamp
- sleep
- count


src/cpp/dtrace/action/host_timestamp.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    m_result = fields[0];
+
+    aiebu::smatch action;
+    if (!aiebu::regex_match(fields[1], action, action_name::action_regex))
+        DTRACE_ERROR("DTRACE_ACTION_INVALID_TOKEN", 
+            "Invalid token: '" << token << "' Expected 'host_timestamp()'");
+
+    m_action_name = action[1];
+}
```
xuhz, line 43: nit:
where do we need this m_action_name?
I saw in other action types, we set the action name as well

xuhz (approved):

src/cpp/dtrace/action/host_timestamp.cpp
AdvaitNaik: m_action_name is a member variable of the base action class and is set in the constructor of each derived action type.
It is typically used in the serialize function to identify or print the action name.

However, since each action type implements its own serialize function, and the action name is always known for each derived class, we can simply hard-code the action name string directly in each serialize implementation instead of storing it in m_action_name.

I will remove m_action_name from the class and avoid redundant member variables.

PR #222, opened 2026-02-05
ensure all columns have the same number of preemption points

HimanshuChoudhary-Xilinx:
Added support to ensure all columns have the same number of preemption points else it may lead to cert hang as all cert synchronize at preemption points

Changes:
Added preempt opcode counting mechanism to track occurrences per column
Implemented verification logic to compare preempt counts across all columns using column 0 as reference
Added error reporting for mismatched preempt opcode counts between columns

### DIFF
--- a/src/cpp/preprocessor/aie2ps/aie2ps_preprocessor.h
+++ b/src/cpp/preprocessor/aie2ps/aie2ps_preprocessor.h
@@ -87,7 +87,7 @@ public:
 
     // Verify PREEMPT opcode count is equal across all columns in the control code.
     // All controllers must have the same number of preemption points to ensure consistent
-    // in cert as cert synchronize preemption points at runtime and missmatch may lead to hang.
+    // in cert, as cert synchronize preemption points at runtime and mismatch may lead to hang.
     // The count from column 0 is used as the reference for comparison.
     if (parser->has_preempt()) {
       auto [success, expected_count, mismatch_col, mismatch_count] = parser->verify_preempt_count();
@@ -97,8 +97,7 @@ public:
           " preempt opcodes (from col 0), but controller " + std::to_string(mismatch_col) +
           " has " + std::to_string(mismatch_count) + " preempt opcodes\n");
       }
-      log_info() << "Preempt opcode count verification passed: " << expected_count
-                 << " preempt opcodes in each of " << parser->get_preempt_col_count() << " controllers" << std::endl;
+      log_info() << "Ctrlcode has " << expected_count << " preemption points\n";
     }
 
     // Verify .target directive matches the -t command line option
--- a/src/cpp/preprocessor/aie2ps/aie2ps_preprocessor.h
+++ b/src/cpp/preprocessor/aie2ps/aie2ps_preprocessor.h
@@ -85,6 +85,22 @@ public:
 
     parser->parse_lines();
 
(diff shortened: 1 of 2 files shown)


src/cpp/preprocessor/asm/asm_parser.h
```diff
@@ -364,10 +374,62 @@ class asm_parser: public std::enable_shared_from_this<asm_parser>
     std::string save_label = "save_" + std::to_string(index);
     std::string restore_label = "restore_" + std::to_string(index);
     m_preempt_labels[group] = {save_label, restore_label};
+    // Increment preempt count for this group/col
+    if (m_current_col >= 0) {
+      auto it = m_col.find(m_current_col);
+      if (it != m_col.end())
```
sonals: When will `it` be empty?

src/cpp/preprocessor/asm/asm_parser.h
```diff
@@ -364,10 +374,62 @@ class asm_parser: public std::enable_shared_from_this<asm_parser>
...
+  bool has_preempt() const {
+    for (const auto& [col, data] : m_col) {
+      if (data.get_preempt_count() > 0)
+        return true;
+    }
+    return false;
+  }
+
+  // Get the number of columns that contain PREEMPT opcodes
+  size_t get_preempt_col_count() const {
```
sonals: Do we need this function as either _all_ columns have preemt or _no_ columns have preempt?

src/cpp/preprocessor/aie2ps/aie2ps_preprocessor.h
```diff
@@ -85,6 +85,22 @@ class aie2ps_preprocessor: public preprocessor
...
+    if (parser->has_preempt()) {
+      auto [success, expected_count, mismatch_col, mismatch_count] = parser->verify_preempt_count();
+      if (!success) {
+        throw error(error::error_code::invalid_asm,
+          "Preempt opcode count mismatch: expected " + std::to_string(expected_count) +
+          " preempt opcodes (from col 0), but controller " + std::to_string(mismatch_col) +
+          " has " + std::to_string(mismatch_count) + " preempt opcodes\n");
+      }
+      log_info() << "Preempt opcode count verification passed: " << expected_count
+                 << " preempt opcodes in each of " << parser->get_preempt_col_count() << " controllers" << std::endl;
```
sonals: We should simplify this message so just say "Ctrlcode has " + std::to_string(expected_count) + " preemption points\n"

sonals (changes requested):
Please make the the minor changes.

src/cpp/preprocessor/asm/asm_parser.h
HimanshuChoudhary-Xilinx: it will never be empty.... but just added check ... removed

src/cpp/preprocessor/asm/asm_parser.h
HimanshuChoudhary-Xilinx: not needed .... was only needed in comment.
simplified comment so no more needed

src/cpp/preprocessor/aie2ps/aie2ps_preprocessor.h
HimanshuChoudhary-Xilinx: done

larry9523 (approved):

PR #233, opened 2026-02-25
Fix dtrace result file creation api deleting handle

AdvaitNaik:

Fix dtrace result file creation API taking ownership of the handle and deleting it on return
Added explicit `create_dtrace_handle() / destroy_dtrace_handle()` APIs to make ownership and lifecycle management explicit for callers.

In multiple run testcase, dtrace handle was freed after the first call and same dtrace handle was reused across multiple runs, which caused SIGSEGV on the second and later calls.

Stopped adopting/deleting the handle in dtrace result file creation API; use a raw pointer so the caller keeps ownership
Added explicit lifecycle APIs: `create_dtrace_handle()` to allocate and initialize context and `destroy_dtrace_handle()` to release it.

### DIFF
--- a/src/cpp/dtrace/dtrace.cpp
+++ b/src/cpp/dtrace/dtrace.cpp
@@ -221,10 +221,8 @@ get_dtrace_result_file(dtrace_handle_t dtrace_handle, const char* result_file)
 {
     try
     {
-        // dtrace handle unique pointer
-        auto handle = std::unique_ptr<dtrace_command_handle>(
-            static_cast<dtrace_command_handle*>(dtrace_handle)
-        );
+        // dtrace handle
+        auto* handle = static_cast<dtrace_command_handle*>(dtrace_handle);
 
         // Initialize the result buffers and memory buffers vector
         std::unordered_map<uint32_t, std::vector<uint32_t>> result_buffers;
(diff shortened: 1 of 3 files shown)


chvamshi-xilinx (approved):

xuhz (approved):

src/cpp/dtrace/dtrace.cpp
```diff
@@ -48,25 +48,19 @@ struct dtrace_command_handle {
...
-        handle->g_control = 
+        // and set the log level
+        handle->g_control =
             std::make_unique<dtrace::control>(script_file, map_data, log_level);
 
-        // Get the number of uC in the script file
-        uint32_t number_uC = handle->g_control->m_control_uC_indices.size();
-        *buffers_length = number_uC;
-
         return static_cast<dtrace_handle_t>(handle.release());
```
rbramand-xilinx, line 64: may be add a comment here explaining why we are calling .release()
// returning raw pointer to caller as its C style interface and its callers responsibility to call destroy handle.

Also document the same in header as well

rbramand-xilinx (approved):

src/cpp/dtrace/dtrace.cpp
chvamshi-xilinx: @AdvaitNaik , Please push it in next PR.
Am merging this change

PR #238, opened 2026-03-06
AIESW-26641: Increase warning level on Windows to w4

HimanshuChoudhary-Xilinx:
added
  /W4           # warning level
  /external:W0  # suppress warnings from system/external include directories


cmake/windows.cmake
```diff
@@ -17,7 +17,8 @@ endif()
 add_compile_options(
   /Zc:__cplusplus
   /WX           # treat warnings as errors
-  /W2           # warning level
+  /W4           # warning level
+  /external:W0  # suppress warnings from system/external include directories
```
stsoe: But aren't you simply silencing warnings here for any headers that are used with SYSTEM, which I see you have added in a few places?

stsoe (commented):
Thank you.  One can argue that with most of the static_casts, the changes are at the surface of the code to preserve existing intergral sizes.   Maybe a deeper dive could be done to change return type of functions or argument types?

Of course static_casts are necessary at some level, and they are an indication that you have thought about the validity of the type change,  but there are maybe too many of them here.

I am concerned about the external:W0 switch.

cmake/windows.cmake
HimanshuChoudhary-Xilinx: removed

HimanshuChoudhary-Xilinx:
> Thank you. One can argue that with most of the static_casts, the changes are at the surface of the code to preserve existing intergral sizes. Maybe a deeper dive could be done to change return type of functions or argument types?
> 
> Of course static_casts are necessary at some level, and they are an indication that you have thought about the validity of the type change, but there are maybe too many of them here.
> 
> I am concerned about the external:W0 switch.

hi @stsoe 
I tried to remove all unnecessary static_cast  but now most of them are vector.size() getting casting to uint32_t which are difficult to avoid

stsoe (approved):

PR #241, opened 2026-03-11
api to expose all savetimestamps line and file in controlcode

HimanshuChoudhary-Xilinx:

### DIFF
--- a/src/cpp/analyzer/transform_manager.cpp
+++ b/src/cpp/analyzer/transform_manager.cpp
@@ -1172,5 +1172,55 @@ get_kernel_instances(const std::string& kernel_name)
   return instances;
 }
 
+std::string
+transform_manager::
+get_dump_section_json()
+{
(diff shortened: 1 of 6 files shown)


chvamshi-xilinx (approved):

chvamshi-xilinx (approved):

larry9523 (commented):
@HimanshuChoudhary-Xilinx  as discussed, let's try to make it generic than a specific savetimestamps opcode.

chvamshi-xilinx (approved):

larry9523 (approved):
I would suggest not introducing a new API op_code (enum), which will be a maintenance burden. Just use the real value of the opcode (in ISA spec) as the interface. This is a special advanced API anyway. The caller should know the ISA and ASM well enough to call this API.

I don't have strong opinion on this though. I approve this PR with comment. :)

chvamshi-xilinx (approved):

PR #253, opened 2026-03-26
limit max section in elf as elfio dont support extended section

HimanshuChoudhary-Xilinx:

### DIFF
--- a/src/cpp/elf/elfwriter.h
+++ b/src/cpp/elf/elfwriter.h
@@ -21,6 +21,7 @@ constexpr int phdr_align = 8;
 constexpr int program_header_static_count = 2;
 constexpr int program_header_dynamic_count = 3;
 constexpr int SHT_CUSTOM_SECTION = ELFIO::SHT_LOUSER + 1;
+constexpr ELFIO::Elf_Word max_sections = 65279; //65280 - 1;
 
 constexpr ELFIO::Elf_Word NT_XRT_UID = 4;
 constexpr ELFIO::Elf_Word NT_XRT_UUID       = 5;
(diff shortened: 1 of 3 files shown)


src/cpp/elf/elfwriter.h
```diff
@@ -20,6 +20,7 @@ constexpr int data_align = 16;
 constexpr int phdr_align = 8;
 constexpr int program_header_static_count = 2;
 constexpr int program_header_dynamic_count = 3;
+constexpr ELFIO::Elf_Word max_sections = 65279; //65280 - 1;
```
larry9523, line 26: maybe add a comment on the limit before extended section is 65279 based on elf spec

src/cpp/elf/elfwriter.h
```diff
@@ -106,6 +107,13 @@ class elf_writer
   void add_group(const std::string& name, const std::vector<uint32_t>& member, ELFIO::Elf_Word info_index);
   uint64_t get_virtual_addr(uint64_t in_prev_virtual_addr, uint64_t in_prev_seg_size);
   uint64_t align_address(uint64_t address);
+  // Helper function to return ELFIO::section* and take section name as input
+  ELFIO::section* add_section_by_name(const std::string& section_name) {
+    if (m_elfio.sections.size() >= max_sections)
```
larry9523: should it be > instead of >=
?

larry9523 (commented):
Besides the comments in line, why do we have a max_pages # limit?

sonals (approved):

larry9523 (approved):

PR #256, opened 2026-03-27
Memory optimization

HimanshuChoudhary-Xilinx:
### Memory Optimizations

**Opt-1 — Zero-copy line iteration in parse_lines()**
asm_parser.cpp

std::istringstream isstr(str) was used to iterate lines. istringstream copies the entire input string into its internal buffer on construction. Replaced with a custom read_next_line lambda that scans the original str buffer directly using find('\n') and assign() — no copy of the file content is ever made.

**Opt-2 — Lazy argument splitting in operation**
asm_parser.h

operation stored std::vector<std::string> m_args, which was eagerly populated by splitting the argument string on every construction. Replaced with std::string m_args_str that stores the raw lowercased argument string as-is. get_args() parses and returns a vector<string> only when a caller actually needs individual tokens. Callers that call it more than once capture the result in a local variable.

**Opt-3 — Inline operation into asm_data**
Files: asm_parser.h, asm_parser.cpp, pager.cpp, aie2_asm_preprocessor_input.cpp

asm_data::m_op was std::shared_ptr<operation>, requiring a separate heap allocation for the control block and the object on every instruction. Changed to a plain inline operation value member — allocated as part of the asm_data struct itself, no separate heap block, no reference counting overhead. All call sites updated from pointer semantics to use get_operation() returning a raw pointer. Null checks removed since the inline member is always valid.

**Opt-4 — Remove m_line from asm_data**
Files: asm_parser.h, asm_parser.cpp

asm_data stored std::string m_line containing the raw assembly text of each instruction. This was used only for debug output. Removed the member entirely. Added get_line() which reconstructs the text on demand by concatenating m_op.get_name() and m_op.get_args_str() — no persistent storage required.

**Opt-5 — Replace m_file string with an intern table index**
asm_parser.h

asm_data stored std::string m_file. Since many instructions share the same filename, Replaced with uint32_t m_file_idx — a 4-byte index into a thread_local std::vector<std::string> intern table (detail::g_filename_table). The table stores each unique filename exactly once. get_file() looks up the string on demand; get_file_idx() returns the raw index for use as a map key prefix. 

**Opt-6 — Use file index in all map key construction**
Files: assembler_state.h, assembler_state.cpp, pager.cpp, report.h, aie2ps_encoder.cpp

It built label/job map keys by concatenating the filename string: get_file() + ":" + labelname. Replaced with std::to_string(get_file_idx()) + ":" + labelname. Keys are now shorter integers ("0:label" instead of "default:label"), and the filename string is never looked up during parsing or encoding. In report.h, Function was changed to store uint32_t m_file_idx instead of std::string m_filename; get_filename() resolves it from the intern table only when writing JSON.

**Opt-7 — Stream debug JSON; never build in-memory tree**
Files: report.h, aie2ps_encoder.cpp

Debug::to_json() constructed a full nlohmann::json object tree covering every instruction in the binary. This built a massive in-memory tree. Additionally, the resulting JSON string was accumulated into section_writer::m_data through repeated std::vector reallocations (geometric doubling). Replaced with write_json(std::ostream&) methods on both Line and Debug that emit JSON tokens directly to the stream. For debug_map.json, output goes straight to std::ofstream. For the .dump ELF section, a thin vec_streambuf shim streams directly into the target std::vector<uint8_t>.

**Opt-8 — Guard debug object creation with dump flag**
aie2ps_encoder.cpp

m_debug.add_function() and m_debug.add_textline() were called unconditionally for every text instruction regardless of the dump mode flag. In disable mode, the resulting Line objects and their opcode string copies were created inside m_debug and immediately wasted. Wrapped both calls in if (m_dump_flag != asm_dump_flag::disable) so in disable mode no Line objects are created and no strings are copied into the debug structure.


src/cpp/common/assembler_state.cpp
```diff
@@ -120,7 +120,7 @@ process(bool makeunique)
       // only used in pager
       // since label are after control code ops we need to record intermediately and resolve it later
       if (!name.compare("apply_offset_57") && makeunique)
-        apply_label_map[data->get_file() + ":" + data->get_operation()->get_args()[0].substr(1)]
+        apply_label_map[std::to_string(data->get_file_idx()) + ":" + data->get_operation()->get_args()[0].substr(1)]
```
sonals: Should the label generation move to its own inline function? Here the label generation algorithm is inline with where it is used.

src/cpp/preprocessor/asm/pager.cpp
```diff
@@ -122,7 +122,7 @@ extractlabels(assembler_state& state, std::shared_ptr<asm_data> token)
       auto lb = arg.substr(1);
       if (state.containscratchpads(lb))
         continue;
-      lb = token->get_file() + ":" + lb;
+      lb = std::to_string(token->get_file_idx()) + ":" + lb;
```
sonals: Please move this logic as a member function of **class asm_data**. This will remove code duplication every where in the code base.

```
std::string asm_data::get_label(const std::string &lbl) {
    return std::to_string(this->get_file_idx()) + ":" + lbl;
}
```

src/cpp/preprocessor/asm/asm_parser.h
```diff
@@ -213,21 +259,33 @@ class asm_data
...
   }
 
   HEADER_ACCESS_GET_SET(code_section, section);
   HEADER_ACCESS_GET_SET(offset_type, size);
   HEADER_ACCESS_GET_SET(pageid_type, pagenum);
   HEADER_ACCESS_GET_SET(uint32_t, linenumber);
-  HEADER_ACCESS_GET_SET(std::string, file);
-  HEADER_ACCESS_GET_SET(std::string, line);
+  const std::string& get_file() const { return detail::lookup_filename(m_file_idx); }
+  void set_file(const std::string& val) { m_file_idx = detail::intern_filename(val); }
```
sonals: Why is set_file() needed?

src/cpp/preprocessor/asm/asm_parser.h
```diff
@@ -213,21 +259,33 @@ class asm_data
...
+  uint32_t get_file_idx() const { return m_file_idx; }
+
+  // Reconstructs the assembly line text from the stored operation on demand.
+  // Returns lowercased text (e.g. "vldr\tr0, [sp, #4]"); no heap copy is stored.
+  std::string get_line() const {
+    const auto& n = m_op.get_name();
+    const auto& a = m_op.get_args_str();
+    if (a.empty()) return n;
+    return n + '\t' + a;
```
sonals: No tabs please, use space.

src/cpp/preprocessor/asm/asm_parser.h
```diff
@@ -213,21 +259,33 @@ class asm_data
...
+  }
+
   bool isLabel() { return m_optype == operation_type::label; }
   bool isOpcode() { return m_optype == operation_type::op; }
   bool isAnnotation() { return m_optype == operation_type::annotation; }
-  std::shared_ptr<operation> get_operation() { return m_op; }
-  void set_operation(std::shared_ptr<operation> op) { m_op = op; }
+  // Returns a raw pointer to the inline operation — always non-null.
+  operation* get_operation() { return &m_op; }
+  const operation* get_operation() const { return &m_op; }
```
sonals: Why pointer? Can the callers get a const reference instead?

src/cpp/preprocessor/asm/asm_parser.h
```diff
@@ -213,21 +259,33 @@ class asm_data
...
+  void set_operation(operation op) { m_op = std::move(op); }
```
sonals: Rename `set_operation()` to `update_operation()`.

sonals (changes requested):
This is a very good change in the design. Please make the suggested changes to polish this further.

sonals:
1. Please add a AIE4/AIE2ps testcase with about 100 MB of ASM operations. We should use it to qualify the builds. The largest one we have is about 1.9 MB which is not sufficient.

2. For command line use of aiebu I would recommend calling the metrics_report() API to show the resource usage.

HimanshuChoudhary-Xilinx:
> 1. Please add a AIE4/AIE2ps testcase with about 100 MB of ASM operations. We should use it to qualify the builds. The largest one we have is about 1.9 MB which is not sufficient.
> 2. For command line use of aiebu I would recommend calling the metrics_report() API to show the resource usage.



> 1. Please add a AIE4/AIE2ps testcase with about 100 MB of ASM operations. We should use it to qualify the builds. The largest one we have is about 1.9 MB which is not sufficient.
> 2. For command line use of aiebu I would recommend calling the metrics_report() API to show the resource usage.

i will add testcase in new pr

sonals (approved):

PR #257, opened 2026-03-30
added const elfio constructor

HimanshuChoudhary-Xilinx:


HimanshuChoudhary-Xilinx:
if we dont use const cast because
error: passing ‘const ELFIO::elfio’ as ‘this’ argument discards qualifiers [-fpermissive]
  427 |   elf->save(stream);

PR #258, opened 2026-03-30
single section changes

HimanshuChoudhary-Xilinx:
 Merge all pages per column into a single ELF section

 For large AIE2PS/AIE4 workloads the legacy per-page encoding produces two ELF sections per page per column:
         .ctrltext.<col>.<page>   (text instructions)
         .ctrldata.<col>.<page>   (buffer descriptors)
With many columns, pages, and kernel instances (config ELF), the total section count exceeds 0xFF00 (65 280) — the ELFIO maximum for a 16-bit section index. This causes ELFIO to produce a corrupt or zero-size ELF.

Merge all pages for a given column into a single contiguous ELF section per column per instance. Each page occupies exactly PAGE_SIZE (8192) bytes inside the section with the layout:

         [page header 16B] 
         [text instructions T_N]
         [buffer descriptors D_N]
         [zero padding to PAGE_SIZE]
The merged section is named .ctrltext.<col> (no page-index suffix), making it structurally unambiguous from the per-page format (.ctrltext.<col>.<page>) and the config ELF group-prefixed format (.ctrltext.<col>.<group_index>).

Spec: https://amd.atlassian.net/wiki/spaces/AIE/pages/1582276490/Single+Section+Per-Column+ELF+Format


sonals:
Please fill the template on the goal of the PR...

src/cpp/elf/aie_elf_constants.h
```diff
@@ -21,10 +21,10 @@ constexpr unsigned char osabi_aie4z        = 0x69;  // 105 - AIE4Z
 // ELF ABI Version values
 // Encoding: upper nibble = major version, lower nibble = minor version
 // ============================================================================
-constexpr unsigned char elf_version_legacy        = 0x02;  // 0.2 - Non-config (all architectures)
-constexpr unsigned char elf_version_legacy_config = 0x03;  // 0.3 - Legacy config (aie2ps/aie4 without .target)
+constexpr unsigned char elf_version_legacy        = 0x03;  // 0.3 - Non-config (all architectures)
```
sayyanna: If the ELF has OS_ABI version as 0x46 and ABI version as 0x03, it matches both the categories: per page based full ELF and merged-page based non/legacy ELF. XRT/elf_loader will fail to identify the correct ELF. Please check

sonals (approved):

sayyanna (approved):

PR #263, opened 2026-04-13
perf optimization: return reference of data part of asm

HimanshuChoudhary-Xilinx:
perf optimization: return reference of data part of asm

Hintmap symbol only in section_asmdata::data  
`col_data::get_label_asmdata_data(label) const` return only data part of asm as hintmap will part of data only.


larry9523 (approved):
Looks reasonable.

So we won't hit this performance issue when there is no preemption hintmap in the asm control code?

sonals (approved):

PR #266, opened 2026-04-15
Minor perf improvements related to number parsing

sonals:
Number parsing was taking up 28% of time for the aie4 multi_graph testcase. Improve speed by removing redundant string operations and making the handler table static.. 

valgrind analysis identified the hotspot.

Low

ctest validated

NA

### DIFF
--- a/src/cpp/common/assembler_state.cpp
+++ b/src/cpp/common/assembler_state.cpp
@@ -4,6 +4,7 @@
 
 #include "utils.h"
 
+#include "regex_wrapper.h"
 #include "aiebu/aiebu_error.h"
 
 #include <functional>
@@ -162,7 +163,7 @@ process(bool makeunique)
  * 4. If string is numeric string: it will convert to decimal
  */
 uint32_t assembler_state::parse_num_arg(const std::string& str) {
-  const std::unordered_map<std::string, std::function<uint32_t(const std::string&)>> handlers = {
+  static const std::unordered_map<std::string, std::function<uint32_t(const std::string&)>> handlers = {
     {"@", [this](const std::string& s) -> uint32_t {
           //If string start with '@': it can be either pad name or label name
           auto key = s.substr(1);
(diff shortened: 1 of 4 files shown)


src/cpp/common/assembler_state.cpp
```diff
@@ -162,7 +163,7 @@ process(bool makeunique)
  * 4. If string is numeric string: it will convert to decimal
  */
 uint32_t assembler_state::parse_num_arg(const std::string& str) {
-  const std::unordered_map<std::string, std::function<uint32_t(const std::string&)>> handlers = {
+  static const std::unordered_map<std::string, std::function<uint32_t(const std::string&)>> handlers = {
```
HimanshuChoudhary-Xilinx: Pipeline is failing because we made 'handlers' as static which uses 'this' pointer.
Function-local static handlers is constructed once (first time the function runs). The lambdas 'this' is fixed at that moment.
Later calls on a different instance still run the same static lambdas, which still use the 'this' from the first call. You read/write the wrong maps and actors.

sonals: I noticed that, moved the table to class definition itself. Will update the PR.

src/cpp/common/assembler_state.cpp
```diff
@@ -202,13 +204,28 @@ uint32_t assembler_state::parse_num_arg(const std::string& str) {
     uint32_t col = std::stoi(str.substr(col_start));
     uint32_t row = std::stoi(str.substr(row_start));
     return (((col & col_mask) << col_shift) | (row & row_mask));
-  } else if (str.rfind("0x") == 0)
+  }
+  */
+  static const regex kTileRe{R"(^tile_(\d+)_(\d+)$)"};
+  smatch m;
+  if (regex_match(str, m, kTileRe) && (m.size() == 3))
+  {
```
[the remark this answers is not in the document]
sonals: Updated the code to check for the letter 't' before running regex

src/cpp/common/assembler_state.cpp
sonals: Moved the table to the class.

HimanshuChoudhary-Xilinx (approved):

PR #268, opened 2026-04-21
Bugfix: Reduce latency when dtrace enabled

svup-xilinx:
The get_action_location_mapping() returned std::unordered_map<uint32_t, uint32_t> by value, meaning every call triggered a full copy of the map allocating new memory and copying all key-value pairs.
when dtrace enabled run.wait() time increased from 2ms to 198ms. This is caused due to calling get_action_location_mapping()
in a loop which is copying the map for every iteration as it returns by value.
Changed the return type to const std::unordered_map<uint32_t, uint32_t>& — a const reference — so the function returns a direct reference to the existing map stored in m_primary_action_location_mapping, with zero copying overhead.
Low
Tested by enabling dtrace observed the run.wait() time improved after the fix.

Latency improved from 198ms to 3 ms when dtrace enabled.
No

### DIFF
--- a/src/cpp/dtrace/pager/pager.cpp
+++ b/src/cpp/dtrace/pager/pager.cpp
@@ -572,7 +572,7 @@ copy_actions(const std::vector<uint32_t>& source, uint32_t offset,
  * This function returns an unordered map that maps action identifiers to their
  * corresponding locations for the given uC index.
  */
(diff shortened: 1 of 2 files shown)


chvamshi-xilinx (approved):

AdvaitNaik (approved):
Great Catch!

xuhz (approved):

PR #272, opened 2026-04-22
Performance optimization 

HimanshuChoudhary-Xilinx:
This change improves hot-path performance in the ASM preprocessor by reducing repeated work when building asm_data rows and when handling certain directives.

1. Filename interning (asm_parser.h): detail::intern_filename no longer scans the whole thread-local filename table on every call. A thread-local std::unordered_map<std::string, uint32_t> maps paths to indices so lookup/insert is amortized O(1) instead of O(n) per intern.
2. One intern per parse (asm_parser.cpp): parse_lines computes parse_file_idx once and passes that index into asm_data for labels and ops, instead of calling intern_filename on every row.
3. Directive fast path (asm_parser.cpp): Lines that start with . but are not directives handled by operate_directive (.long, .align, .eof, .eop) skip the directive regex path, avoiding unnecessary work on those prefixes.

**For aie4_compile_time**
**1. linux --> time reduces from 140 sec to 96 sec
2. windows --> time reduces from 289.45 sec to 185 sec**

### DIFF
--- a/src/cpp/preprocessor/asm/asm_parser.cpp
+++ b/src/cpp/preprocessor/asm/asm_parser.cpp
@@ -48,33 +48,23 @@ namespace aiebu {
 
 namespace {
 
-// Order must match asm_directive_id and asm_parser::parse_lines() handler registration.
-constexpr const char* k_asm_directive_pattern_literals[] = {
-  ".attach_to_group",
-  ".aie_row_topology",
-  ".include",
-  ".endl",
-  ".setpad",
-  ".section",
-  ".partition",
-  ".target",
+// One capture per directive token; capture 13 is optional args. Handled directives are captures 1–8
+// (same eight as asm_directive_id / directive_list). Captures 9–12 are recognized but not dispatched here.
+// Order of groups 1–8 must match k_alt_capture_to_handler_idx below.
+const regex DIRECTIVE_ALT_RE(
+    R"(^(?:(\.attach_to_group)|(\.include)|(\.endl)|(\.setpad)|(\.section)|(\.partition)|(\.target)|(\.aie_row_topology)|(\.align)|(\.long)|(\.eop))(?:[ \t]+(.+))?$)");
+
+// Map DIRECTIVE_ALT_RE capture index 1..8 -> asm_parser::directive_list index (asm_directive_id order).
+constexpr std::array<std::size_t, 8> k_alt_capture_to_handler_idx = {
+    static_cast<std::size_t>(asm_directive_id::attach_to_group),
+    static_cast<std::size_t>(asm_directive_id::include),
+    static_cast<std::size_t>(asm_directive_id::endl),
+    static_cast<std::size_t>(asm_directive_id::setpad),
+    static_cast<std::size_t>(asm_directive_id::section),
+    static_cast<std::size_t>(asm_directive_id::partition),
+    static_cast<std::size_t>(asm_directive_id::target),
+    static_cast<std::size_t>(asm_directive_id::aie_row_topology),
 };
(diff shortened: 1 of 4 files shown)


src/cpp/preprocessor/asm/asm_parser.cpp
```diff
@@ -202,8 +203,12 @@ parse_lines(const std::vector<char>& data, std::string& file)
...
     smatch sm;
 
-    // Check for Directive (starts with .) - only check regex if prefix matches
-    if (line[0] == '.' && operate_directive(line))
+    if (line[0] == '.' &&
+        line.substr(0,5).compare(".long") != 0 &&
+        line.substr(0,6).compare(".align") != 0 &&
+        line.substr(0,4).compare(".eof") != 0 &&
+        line.substr(0,4).compare(".eop") != 0 &&
+        operate_directive(line))
```
larry9523: maybe we can use
line.size() >= 5 && line.compare(0, 5, ".long") == 0 to avoid extra allocation on hot path

larry9523 (commented):
Can we also update the PR title to be more informative?

src/cpp/preprocessor/asm/asm_parser.cpp
sonals: Instead of trying to match with all the non consequential directives (.long, .align, .eof, .eop) sequentially and then doing another look up via directive_list[] we should build a comprehensive regex with all the directives explicitly instantiated and then use the regex_match() to match the directive token in one shot. The directive_list[] then becomes an array indexed by integer/enum instead of std::map keyed by std::string. This also has the advantage of having a single table of all directives.

Snippets of suggestions from Cursor:

const static std::regex DIRECTIVE_ALT_RE(
        R"(^(?:(\.attach_to_group)|(\.include)|(\.endl)|(\.setpad)|(\.section)|(\.partition)|(\.target)|(\.aie_row_topology)|(\.align)|(\.long)|(\.eof)|(\.eop))(?:[ \t]+(.+))?$)");
    
    smatch m;
    if (!regex_match(line, m, DIRECTIVE_ALT_RE) || m.size() < 14U)
      return false;

    size_t idx = NUM_DIRECTIVE_HANDLERS;
    if (m[1].matched)       idx = 0U;
    else if (m[2].matched)   idx = 1U;
    else if (m[3].matched)   idx = 2U;
    else if (m[4].matched)   idx = 3U;
    else if (m[5].matched)   idx = 4U;
    else if (m[6].matched)   idx = 5U;
    else if (m[7].matched)   idx = 6U;
    else if (m[8].matched)   idx = 7U;
    ...
    if (idx >= NUM_DIRECTIVE_HANDLERS)
      return false;

    const auto& op = directive_list[idx];
    if (!op)
      return false;
   ...

src/cpp/preprocessor/asm/asm_parser.cpp
```diff
@@ -46,6 +46,52 @@ static inline int aiebu_clz(unsigned int x)      { return __builtin_clz(x);
...
+} // namespace
+
+bool
+asm_parser::
+operate_directive(const std::string& line)
+{
+  smatch sm;
+  if (!regex_match(line, sm, k_known_asm_directive_re))
+    return false;
+  const std::size_t idx = index_for_known_directive_name(sm[1].str());
```
sonals: We should not do another series of string comparison here. sm already tells you what it has matched --  sm[idx].match will be true for a specific idx. That idx can be directly used to execute directive_list[idx]->operate()

HimanshuChoudhary-Xilinx: got it ... fixed

sonals (approved):

PR #279, opened 2026-04-27
per optimization

HimanshuChoudhary-Xilinx:
Improvement on multiple hot path and windows build flag changes:
 
1. MSVC / Windows CMake (cmake/windows.cmake)
Release: Explicit /O2 and /Ot (fast code vs size), each in its own $<CONFIG:Release> generator expression so CMake does not misparse a single string with ;.
_ITERATOR_DEBUG_LEVEL=0 in Release to avoid checked-iterator overhead in optimized builds.

2. Pager (pager.cpp / pager.h) - reduced paging time  ~31s → ~8s in linux
    a. Out-of-order opcodes: Replaced repeated std::find on a small opcode list with opcode_is_out_of_order_list() (string compares for load_pdi, preempt, load_cores). The old pager::OOO vector + std::find per argument is gone after cleanup.
    b. Label extraction: extractlabels / recursive label gathering no longer builds intermediate vectors and merges them with union_of_lists_inorder on every step (which implied repeated linear deduplication). The new extractlabels_accum walks arguments once per logical visit, uses an std::unordered_set<std::string> for seen, and appends to the output list only on first sight of a label. Dependent labels from dependent_labelmap are merged with the same dedup rules.
    c. External labels: extract_externallabels_accum mirrors the OOO opcode rules with set-based dedup instead of union_of_lists_inorder for single-element vectors.
    d. extractjobsandlabels: Reuses one dependent_labelmap reference, maintains labels_seen / external_seen sets across the job loop so work is not duplicated when the same label appears from multiple tokens in the same page extraction.

3. assembler_state (assembler_state.cpp / .h)
 a. Single pass over m_data: Loop uses const auto& instead of copying shared_ptrs each iteration.
 b. std::map inserts: Duplicate checks for labels and jobs use lower_bound + key equality and emplace_hint instead of find then operator[], which is cheaper when keys are roughly monotonic in the stream.
 c. ISA lookup: Uses m_isa->find(name) once and isa_it->second instead of count + operator[].
 d. Encoded size: Uses isa_op::encoded_size_in_text directly instead of going through serializer(...)->size for the common path (aligns with new sizing logic in ops).
 e. local_barrier / launch_job: Rely on operator[] on std::map for vectors instead of explicit find + empty-vector insert branches where safe.
 f. .eop naming: gen_eop_name is called once per .eop and reused for map key and push_back.
 
4. ELF writer (elfwriter.cpp / elfwriter.h, derived writers)
  a. m_section_by_name: std::unordered_map<std::string, ELFIO::section*> caches section pointers because ELFIO::sections[name] scans all sections O(n).
  b. add_section_by_name: Registers new sections in the map.
  c. lookup_section: Map hit returns immediately; on miss, falls back to ELFIO lookup and inserts into the map.
  e. resync_section_name_map: Rebuilds the map from ELFIO’s section container; iterates with for (auto& s : m_elfio.sections) and uses s->get_name() / s.get() because the range yields std::unique_ptr<section>& (not raw section&).
max_sections: Guard uses >= max_sections so the limit cannot be exceeded by one (off-by-one style fix in fix).
  f. aie2_blob_elfwriter.h / aie2ps_elfwriter.h: Group section assembly adjusted for the writer base (comments about prepend without O(n) insert-at-begin); fix drops an extra grouped.reserve that was tied to that pattern.
  g. add_dynsym_section: fix removes hash.reserve(...) on the symbol dedup map.

5. ASM parser: .long vs directives (asm_parser.cpp)
Directive handling used:
if (line[0] == '.' && line.substr(0,5).compare(".long") && operate_directive(line))

So when the first five characters are .long, compare returns 0 (false) and operate_directive is not invoked for that path—.long is left to the normal opcode / line pipeline.

### Base timing
Linux:        aie4_compile_time .........................   Passed   91.65 sec
Windows: aie4_compile_time .........................   Passed  219.99 sec
### 1. optimization + text -j 4:
Linux:        aie4_compile_time .........................   Passed   43.43 sec
Windows: aie4_compile_time .........................   Passed  103.42 sec
https://github.com/Xilinx/aiebu/actions/runs/24987849232/job/73165588219?pr=279

### 2. Optimization + test -j 1
Linux:       aie4_compile_time .........................   Passed   38.38 sec
Wnidows: aie4_compile_time .........................   Passed   73.61 sec
https://github.com/Xilinx/aiebu/actions/runs/24990069727/job/73173179055?pr=279

### 3. Optimization + test -j 4 + line.sutstr(0,5).compare(".long")
Linux:       aie4_compile_time .........................   Passed   40.90 sec
Windows: aie4_compile_time .........................   Passed   93.70 sec
https://github.com/Xilinx/aiebu/actions/runs/24997939800/job/73199806776?pr=279

### 4. Optimization + test -j 1 + line.sutstr(0,5).compare(".long")
Linux:        aie4_compile_time .........................   Passed   35.34 sec
Windows: aie4_compile_time .........................   Passed   62.70 sec
https://github.com/Xilinx/aiebu/actions/runs/24992903754/job/73182603051?pr=279

### DIFF
--- a/src/cpp/elf/aie2ps/aie2ps_elfwriter.h
+++ b/src/cpp/elf/aie2ps/aie2ps_elfwriter.h
@@ -48,9 +48,11 @@ public:
        {
          auto instance_index = add_symtab_section(iname, kernel_index);
          std::vector<uint32_t> group_data = process_common_helper(instance, get_section_prefix(index));
-         // first word is GRP_COMDAT
-         group_data.insert(group_data.begin(), 1);
-         add_group(get_group_name(index), group_data, instance_index);
+         // first word is GRP_COMDAT (prepend without O(n) insert-at-begin)
+         std::vector<uint32_t> grouped;
+         grouped.push_back(1);
+         grouped.insert(grouped.end(), group_data.begin(), group_data.end());
+         add_group(get_group_name(index), grouped, instance_index);
          index++;
        }
     }
@@ -61,7 +63,8 @@ public:
     std::memcpy(configuration_vec.data(), &col, sizeof(uint32_t));
 
     add_note(NT_XRT_PARTITION_SIZE, xrt_configuration, configuration_vec);
-    return finalize();
+    auto result = finalize();
+    return result;
   }
 };
 }
--- a/cmake/windows.cmake
+++ b/cmake/windows.cmake
@@ -28,6 +28,15 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
   add_compile_options(/Qspectre)  # compile with the Spectre mitigations switch
 endif()
 
+# Release: explicit max-speed opts (/O2; /Ot favors fast code over smaller code) and
+# no checked STL iterators. Base CMAKE_CXX_FLAGS_RELEASE from CMake is /O2 /Ob2 /DNDEBUG;
+# we restate /O2 here so Release tuning stays explicit in this file.
+# NOTE: Put /O2 and /Ot in separate genexes — a single "$<...:/O2;/Ot>" breaks because ';'
+# splits the generator-expression parse and can leak "$<1:/O2" onto the cl command line.
+add_compile_options("$<$<STREQUAL:$<CONFIG>,Release>:/O2>")
+add_compile_options("$<$<STREQUAL:$<CONFIG>,Release>:/Ot>")
+add_compile_definitions($<$<STREQUAL:$<CONFIG>,Release>:_ITERATOR_DEBUG_LEVEL=0>)
+
 add_link_options(
   /DEBUG      # instruct linker to create debugging info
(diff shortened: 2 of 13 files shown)


sonals (approved):

src/cpp/elf/aie2ps/aie2ps_elfwriter.h
```diff
@@ -61,7 +63,8 @@ class aie2ps_config_elf_writer: public elf_writer
     std::memcpy(configuration_vec.data(), &col, sizeof(uint32_t));
 
     add_note(NT_XRT_PARTITION_SIZE, xrt_configuration, configuration_vec);
-    return finalize();
+    auto result = finalize();
+    return result;
```
larry9523, line 67: same as
return finalize();
?

larry9523 (approved):

PR #287, opened 2026-05-13
update save/restore preemption asm annotation points

AdvaitNaik:

update save/restore preemption asm annotation id for simplicity
make annotation id file-scoped (only 0/1 per file)


sonals (commented):
Shouldn't the tables defined in [src/cpp/preprocessor/aie4/aie4_save_restore_map_prebuilt.h](https://github.com/Xilinx/aiebu/pull/287/changes#diff-0a670901d669d350a7419bf67a14b90b7ba409ece3c7cd3c97d37071d5006291) be const?

xuhz (approved):

PR #290, opened 2026-05-17
load_pdi, load_cores, load_cores_cp, preempt, start_cond_job_preempt check

HimanshuChoudhary-Xilinx:

https://amd.atlassian.net/wiki/spaces/AIE/pages/1687216605/load_pdi+load_cores+load_cores_cp+preempt+start_cond_job_preempt+check+in+aiebu

#### load_pdi
```
Restriction (spec)                                              aiebu check required?         Test
--------------------------------------------------------------------------------------------------------------
Each PDI address must be unique within control ELF              Yes                           aie4_check_dup_load_pdi_addr
Equal count across all UCs in multi-UC config                  Yes                           aie4_check_load_pdi_count_mismatch
PDI id uniqueness (shared across ELFs)                        No — up to user               —
```

#### load_cores
```
Restriction (spec)                                              aiebu check required?         Test
--------------------------------------------------------------------------------------------------------------
Each core-ELF address must be unique                           Yes (same as load_pdi)        aie4_check_dup_load_cores_addr
Equal count across all UCs                                    Yes (same as load_pdi)        aie4_check_load_cores_count_mismatch
```

#### load_cores_cp
```
Restriction (spec)                                              aiebu check required?         Test
--------------------------------------------------------------------------------------------------------------
Equal count across all UCs                                    Yes                           aie4_check_load_cores_cp_count_mismatch
Control code must fit in one page                             No — hardware/linker          —
```

#### preempt
```
Restriction (spec)                                              aiebu check required?         Test
--------------------------------------------------------------------------------------------------------------
At least one load_pdi must exist                              Yes                           aie4_check_preempt_no_load_pdi
Equal count across all UCs                                    Yes                           —
```

#### start_cond_job_preempt
```
Restriction (spec)                                              aiebu check required?         Test
--------------------------------------------------------------------------------------------------------------
Must appear after at least one preempt                         Yes                           aie4_check_start_cond_before_preempt
Equal count across all UCs                                    Yes                           aie4_check_start_cond_count_mismatch
```


src/cpp/preprocessor/aie2ps/aie2ps_preprocessor.h
```diff
@@ -101,6 +101,52 @@ class aie2ps_preprocessor: public preprocessor
...
+
+    // Verify load_pdi opcode count is equal across all columns in multi-UC control code.
+    {
+      auto [ok, exp, col, got] = parser->verify_load_pdi_count();
+      if (!ok)
+        throw error(error::error_code::invalid_asm,
+          "load_pdi opcode count mismatch: expected " + std::to_string(exp) +
+          " (from col 0), but controller " + std::to_string(col) +
+          " has " + std::to_string(got) + " load_pdi opcodes\n");
+    }
```
sonals, line 120: Simplify the message:

"load_pdi opcode count mismatch: std::to_string(exp) + " for col 0" + "  but " + std::to_string(got) + "for col " + std::to_string(col)

src/cpp/preprocessor/aie2ps/aie2ps_preprocessor.h
```diff
@@ -101,6 +101,52 @@ class aie2ps_preprocessor: public preprocessor
...
+
+    // Verify start_cond_job_preempt opcode count is equal across all columns.
+    {
+      auto [ok, exp, col, got] = parser->verify_start_cond_job_preempt_count();
+      if (!ok)
+        throw error(error::error_code::invalid_asm,
+          "start_cond_job_preempt opcode count mismatch: expected " + std::to_string(exp) +
+          " (from col 0), but controller " + std::to_string(col) +
+          " has " + std::to_string(got) + " start_cond_job_preempt opcodes\n");
     }
```
sonals, line 157: Simplify as suggested above

src/cpp/preprocessor/asm/asm_parser.cpp
```diff
@@ -388,6 +388,63 @@ parse_lines(const std::vector<char>& data, std::string& file)
...
       }
+      // Per-opcode parse-time checks and counter updates.
+      // These run after the preempt block so that m_preempt_count is already
+      // incremented when we reach start_cond_job_preempt.
+      else if (op_name == "load_pdi" || op_name == "load_cores" ||
+               op_name == "load_cores_cp" || op_name == "start_cond_job_preempt") {
+        // col is common to all four opcodes; compute once.
+        int col = current_col();
+
+        if (op_name == "load_pdi" || op_name == "load_cores") {
```
sonals: Why not use std::regex here to extract the label for both opcodes?

test/aie4-ctrlcode/opcode_checks/dup_load_cores_addr.asm
```diff
@@ -0,0 +1,28 @@
+; Negative test: two LOAD_CORES instructions in the same column reference the
+; same @label (same core-elf host address).  Each load_cores within a control
+; code elf must point to a distinct address.
+; Expected error: "load_cores location '@cores0' is not unique in column 0"
+.target aie4
+.partition 1column
+.attach_to_group 0
+START_JOB 0
+	LOAD_PDI 0, @pdi0
```
xuhz, line 9: each load_pdi, load_cores should take one job.
putting multiple load_cores opcode in one job is meaningless, but it is valid -- real loading happens after current job, so multiple load_cores opcode results in we remember just the id of 2nd one, and load only 2nd one as well.
we should add one extra check -- in one job, there should be at most one load_pdi, load_cores, load_cores_cp, preempt.

test/aie4-ctrlcode/opcode_checks/dup_load_pdi_addr.asm
```diff
@@ -0,0 +1,21 @@
+; Negative test: two LOAD_PDI instructions in the same column reference the same
+; @label (i.e. the same PDI host address).  Each load_pdi within a control code
+; elf must point to a distinct address.
+; Expected error: "load_pdi location '@pdi0' is not unique in column 0"
+.target aie4
+.partition 1column
+.attach_to_group 0
+START_JOB 0
```
xuhz, line 8: same

test/aie4-ctrlcode/opcode_checks/load_cores_count_mismatch.asm
```diff
@@ -0,0 +1,35 @@
+; Negative test: multi-UC control code where load_cores counts differ.
+; col 0: 1 LOAD_CORES   col 1: 0 LOAD_CORES
+; Expected error: "load_cores opcode count mismatch"
+.target aie4
+.partition 2column
+.attach_to_group 0
+START_JOB 0
```
xuhz, line 7: same to all remaining test files

xuhz (changes requested):

src/cpp/preprocessor/asm/asm_parser.h
```diff
@@ -632,6 +669,57 @@ class asm_parser: public std::enable_shared_from_this<asm_parser>
     return false;
   }
 
+  // Generic helper: verify all columns have the same count for a given opcode.
+  // count_getter must return a uint32_t count from a col_data reference.
+  // Returns {success, expected_count, mismatched_col, mismatched_count}.
+  template<typename CountGetter>
+  std::tuple<bool, uint32_t, int, uint32_t>
```
sonals: counters should be `int` or `unsigned int`. Why are they `uint32_t`?

src/cpp/preprocessor/aie2ps/aie2ps_preprocessor.h
```diff
@@ -101,6 +101,52 @@ class aie2ps_preprocessor: public preprocessor
...
+    }
+
+    // Verify load_cores_cp opcode count is equal across all columns.
+    {
+      auto [ok, exp, col, got] = parser->verify_load_cores_cp_count();
+      if (!ok)
+        throw error(error::error_code::invalid_asm,
+          "load_cores_cp opcode count mismatch: " + std::to_string(exp) +
+          " for col 0, but " + std::to_string(got) +
+          " for col " + std::to_string(col) + "\n");
```
sonals: Please use lower case **column** consistently in all messages. I see col, column, Column, etc...

src/cpp/preprocessor/asm/asm_parser.cpp
```diff
@@ -388,6 +390,53 @@ parse_lines(const std::vector<char>& data, std::string& file)
...
+
+          // Extract the @label at argument index 1 (<id>, @<label>).
+          // Compiled once (static); regex_search finds the first ", @<label>" token.
+          static const regex LOAD_LABEL_RE(",\\s*(@[a-zA-Z0-9_]+)");
+          std::string label_arg;
+          smatch lm;
+          if (regex_search(arg_str, lm, LOAD_LABEL_RE))
+            label_arg = lm[1].str();
+          // Enforce uniqueness of the PDI / core-elf address within this column.
+          if (!label_arg.empty()) {
```
sonals: When will this be empty()? If empty() is user error then we should check for that first and bail out if the check fails.

src/cpp/preprocessor/asm/asm_parser.cpp
```diff
@@ -388,6 +390,53 @@ parse_lines(const std::vector<char>& data, std::string& file)
 
         line = op_name + "\t" + arg_str;
       }
+      // Per-opcode parse-time checks and counter updates.
+      // These run after the preempt block so that m_preempt_count is already
+      // incremented when we reach start_cond_job_preempt.
+      else if (regex_match(op_name, sm, LOAD_OR_PREEMPT_COND_RE)) {
```
sonals, line 440: This and the previous "preempt" check should be moved to two helper functions that are called from parse_lines().

src/cpp/preprocessor/asm/asm_parser.h
```diff
@@ -390,6 +391,16 @@ class col_data
   // Counter for PREEMPT opcodes in this column's control code.
   // Used to verify all columns have the same number of preempt points.
   uint32_t m_preempt_count = 0;
+  // Counters for load_pdi, load_cores, load_cores_cp, and start_cond_job_preempt opcodes.
+  // Used to verify all columns have the same count in multi-UC configurations.
+  uint32_t m_load_pdi_count = 0;
+  uint32_t m_load_cores_count = 0;
+  uint32_t m_load_cores_cp_count = 0;
+  uint32_t m_start_cond_job_preempt_count = 0;
```
sonals: These should be `unsigned int`

sonals (changes requested):
See the details:

1. Some code refactoring for the existing and new rules check
2. Message consistency
3. uint32_t vs unsigned int

xuhz (commented):
one more check to add:
if there is load_cores or load_cores_cp, there must be load_pdi 
i already updated the confluence page for this.

src/cpp/preprocessor/asm/asm_parser.h
HimanshuChoudhary-Xilinx: Hi @sonals 
for machine 32/64bit machine both unsigned int or uint32_t are of 32bit 
uint32_t make platform independent... always 32bit

src/cpp/preprocessor/asm/asm_parser.cpp
HimanshuChoudhary-Xilinx: it will never be... removed the check

src/cpp/preprocessor/asm/asm_parser.h
HimanshuChoudhary-Xilinx: same as above

src/cpp/preprocessor/asm/asm_parser.h
sonals: That is not the point. uint32_t is to refer to 32-bit HW registers in our case, AIE registers. int is native C/C++ type and should be used for control flow in the application.

src/cpp/preprocessor/asm/asm_parser.cpp
```diff
@@ -217,6 +217,120 @@ parse_lines()
...
+handle_preempt_opcode(std::string& arg_str, std::string& line)
+{
+  if (get_target_type() == "aie2ps")
+    throw error(error::error_code::internal_error, "PREEMPT opcode is not supported for aie2ps target");
+
+  int current_group = current_col();
+  record_preempt_label(current_group);
+
+  std::vector<std::string> args;
+  if (!arg_str.empty()) {
```
sonals, line 232: Please use std::regex to parse the preempt operation and obtain the arguments.

src/cpp/preprocessor/asm/asm_parser.h
HimanshuChoudhary-Xilinx: done

src/cpp/preprocessor/asm/asm_parser.h
HimanshuChoudhary-Xilinx: done

sonals (approved):

PR #291, opened 2026-05-19
AIESW-32720: aiebu enhancement to check for duplicate file names

HimanshuChoudhary-Xilinx:
Certain testcases seem to use the same asm file name multiple times causing confusion during debug. aiebu should be enhanced to catch duplicate file name situation during compile and provide a proper error message.


xuhz (commented):
the case we experienced was,
.include foo1_folder/bar.asm
.include foo2_folder/bar.asm

functionality wise, these 2 bar.asm files are different, so everything is working fine.
but when we declare dtrace probes, we use only the bar.asm instead of the full path because in real world, the path may be very long, and very inconvenient to use otherwise.

in this case, dtrace will not work if we declare probes in bar.asm.

I would like aiebu has at least warning, or even error for this case.

the current PR seems only check the uniqueness of the full path

src/cpp/preprocessor/asm/asm_parser.h
```diff
@@ -478,6 +478,7 @@ class asm_parser: public std::enable_shared_from_this<asm_parser>
   std::map<int, std::vector<std::string>> m_preempt_hintmaps;  // group -> vector of hintmap_labels (multiple PREEMPT opcodes per group)
   std::map<std::string, std::pair<std::string, std::string>> m_hintmap_labels;  // hintmap_label -> (save_label, restore_label)
   std::set<int> m_preempt_without_hintmap;  // groups that have PREEMPT opcodes without hintmaps
+  std::set<std::string> m_seen_files;  // tracks included file names to detect duplicates
```
sonals: Don't we already have file index table? Why do we need this additional table?

HimanshuChoudhary-Xilinx: yes .. that table was common bet all instance and kernel in config .... 
changes

xuhz (approved):

src/cpp/preprocessor/asm/asm_parser.h
```diff
@@ -294,7 +297,9 @@ class asm_data
   HEADER_ACCESS_GET_SET(offset_type, size);
   HEADER_ACCESS_GET_SET(pageid_type, pagenum);
   HEADER_ACCESS_GET_SET(uint32_t, linenumber);
-  const std::string& get_file() const { return detail::lookup_filename(m_file_idx); }
+  const std::string& get_file(const detail::filename_table& table) const {
+    return table.lookup_filename(m_file_idx);
```
sonals, line 302: Shouldn't this be an operation on `filename_table `itself?
`filename_table` and `file index` should be unified so there is a single database.

HimanshuChoudhary-Xilinx: we have one database for filename and its index for one instance. 
each opcode in asm record file index which when needed can be converted to filename from database lookup 
i see a point to merge to one as it has redundant data but we maintain 2 just for faster lookup  
```
class filename_table {
  std::vector<std::string> m_table;
  std::unordered_map<std::string, uint32_t> m_index;
```

sonals: It would be best to add a note that the redundant data should be removed in future.

HimanshuChoudhary-Xilinx: sure

PR #293, opened 2026-05-26
update isa spec docs

xuhz:
add detailed explanation of some of the opcodes in isa spec

### DIFF
--- a/specification/aie2ps/isa-spec.html
+++ b/specification/aie2ps/isa-spec.html
@@ -792,8 +792,8 @@ In case the number of TCTs haven't all been arrived yet when this operation is e
 the current job will be blocked until all the TCT arrive. The runtime will proceed with
 execution of other (non-blocked) jobs during that time.
 Use this operation only for tasks which have TCTs enabled and where the TCTs are
-routed to arrive at the uC executing this operation.
-Note that tasks can be enqueued from any jobs, but only one job can call <code>WAIT_TCTS</code>. And
+routed to arrive at the uC executing this operation.</p>
+<p>Note that tasks can be enqueued from any jobs, but only one job can call <code>WAIT_TCTS</code>. And
 waiting for more tcts than enqueued will hang the job forever</p>
 <h2 id="end_job-0x07">END_JOB (0x07)</h2>
 <p>Indicates the end of the current job.</p>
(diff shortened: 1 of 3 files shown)


specification/aie2ps/isa-spec.html
```diff
@@ -1179,6 +1179,7 @@ <h2 id="remote_barrier-0x12">REMOTE_BARRIER (0x12)</h2>
 remote barrier, only one job in a column can participate in. Example:</p>
 <pre><code>REMOTE_BARRIER      $rb0, 0xC
 </code></pre>
+<p>Note that it is not necessary all prior jobs have to complete before the barrier matures.</p>
```
sonals: `Note that it is not necessary for all prior jobs to complete before the barrier matures.`

specification/aie2ps/isa-spec.html
```diff
@@ -1378,6 +1379,16 @@ <h2 id="preempt-0x19">PREEMPT (0x19)</h2>
 for each preemption point id, the control code of each uc should have this opcode with same <code>id</code>.
 This opcode should take one whole job but the job can share page with other jobs. All following jobs in same page don't start
 before this preemption job is done</p>
+<p>Note:</p>
+<ol>
+<li>
+<p>Just like REMOTE_BARRIER, the barrier sync done by CERT doesn't require prior jobs to complete</p>
```
sonals: `Like REMOTE_BARRIER handling, CERT does not require all prior jobs to complete when executing PREEMPT.`

Can you clarify if CERT will wait for all previous running jobs to complete before it completes PREEMPT? Or can we go past PREEMPT opcode while previous jobs are still running.

specification/aie2ps/isa-spec.html
```diff
@@ -1439,6 +1462,7 @@ <h2 id="load_cores-0x04">LOAD_CORES (0x04)</h2>
 </tbody>
 </table>
 <p>This essentially is same to LOAD_PDI except that CERT will save the elf info for aie cores to different location than the pdi so that cert can do recovery of both during preemption</p>
+<p>Note that same to LOAD_PDI, reset/unreset, enable/disable core is required to load new core elf. This should be part of generated control code</p>
```
sonals: s/same to/like/

sonals (changes requested):
Please refine the newly added statement with co-pilot.

sonals (approved):

PR #309, opened 2026-06-18
Add data driven AIE4 decode opcode testcase partially generated by Cursor

sonals:
1. Added a JSON based AIE4 decode testcase
2. Early exit if an invalid kernel name is passed to decode 
3. Do not release aiebu_debug.h as public API since the file has dependency on ELFIO itself


test/aie4-ctrlcode/decode_opcode/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+  message(FATAL_ERROR
+    "${DECODE_OPCODE_TEST} requires Boost 1.75 or later (Boost.JSON). "
+    "Found Boost ${Boost_VERSION_STRING}.")
+endif()
+find_package(Boost 1.75.0 REQUIRED COMPONENTS json)
+target_link_libraries(${DECODE_OPCODE_TEST} PRIVATE Boost::json)
+
+target_include_directories(${DECODE_OPCODE_TEST} PRIVATE
+  ${AIEBU_SOURCE_DIR}/src/cpp/include
+  ${AIEBU_SOURCE_DIR}/src/cpp/ELFIO
```
sayyanna: I think the testcase will fail when built with XRT because XRT uses a different ELFIO. Below needs to be used instead.
${AIEBU_ELFIO_SRC_DIR}

sonals: Thanks, updated.

PR #310, opened 2026-06-24
dynamic tracing compiler result buffer API refactor

AdvaitNaik:

Eliminate nlohmann JSON library dependency from aiebu's dtrace public API. Simplify `dtrace_result_buffer()` by using standard string return value semantics.

No bug fix. Design improvement - aiebu was exposing nlohmann::ordered_json in dtrace public headers, forcing all clients to depend on that library.

Changed `get_dtrace_result_buffer(handle)` to returns JSON as string directly.

Low

 Verifed dtrace buffering works end-to-end in XRT on windows medusa board and simnow baremetal flow
 
https://amd.atlassian.net/wiki/spaces/AIE/pages/779183759/cert+dynamic+tracing+dtrace+compiler+APIs

### DIFF
--- a/src/cpp/dtrace/dtrace.cpp
+++ b/src/cpp/dtrace/dtrace.cpp
@@ -286,10 +286,8 @@ get_dtrace_result_file(dtrace_handle_t dtrace_handle, const std::string& result_
     }
 }
 
-/*
-void
-update_dtrace_result_buffer(dtrace_handle_t dtrace_handle, const std::string& result_key,
-    nlohmann::ordered_json& result_buffer)
+std::string
+get_dtrace_result_buffer(dtrace_handle_t dtrace_handle)
 {
     try
     {
@@ -328,10 +326,9 @@ update_dtrace_result_buffer(dtrace_handle_t dtrace_handle, const std::string& re
             mem_buffers[uC_index] = std::move(mem_buffer);
         }
 
-        // Update the result buffer with the given result key and result buffer data
+        // Create the result buffer as JSON object
         nlohmann::ordered_json result_json = nlohmann::ordered_json::object();
         handle->g_control->create_result_buffer(result_buffers, mem_buffers, result_json);
-        result_buffer[result_key] = std::move(result_json);
 
         // Iterate over and copy back modified buffers for each uC in global map
         for (const auto& [uC_index, l_dtrace_buffer_info] : handle->g_dtrace_buffer_info_map)
@@ -351,13 +348,16 @@ update_dtrace_result_buffer(dtrace_handle_t dtrace_handle, const std::string& re
                 buffer.size() * sizeof(uint32_t)
             );
         }
+
+        // Return the result buffer as JSON string
+        return result_json.dump();
     }
     catch (const std::exception& e)
     {
         std::cerr << e.what();
+        return "null";
     }
 }
-*/
 
 void
 destroy_dtrace_handle(dtrace_handle_t dtrace_handle)
--- a/src/cpp/dtrace/dtrace.cpp
+++ b/src/cpp/dtrace/dtrace.cpp
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 // Copyright (C) 2024-2026 Advanced Micro Devices, Inc. All rights reserved.
 
-// This file implements the dtrace public APIs for creating dtrace control buffer, 
+// This file implements the dtrace public APIs for creating dtrace control buffer,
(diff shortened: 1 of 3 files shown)


src/cpp/dtrace/dtrace.cpp
```diff
@@ -351,13 +348,16 @@ update_dtrace_result_buffer(dtrace_handle_t dtrace_handle, const std::string& re
...
         }
+
+        // Return the result buffer as JSON string
+        return result_json.dump();
     }
     catch (const std::exception& e)
     {
         std::cerr << e.what();
+        return "";
     }
```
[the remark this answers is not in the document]
AdvaitNaik: done

xuhz (approved):

src/cpp/dtrace/dtrace.h
stsoe: Update

src/cpp/dtrace/dtrace.h
```diff
@@ -5,7 +5,6 @@
 #define TRACE_H
 
 // This header file contains the public APIs for creating control buffer, memory buffer, and result file
-// #include "json/nlohmann/json.hpp"
 #include "utils.h"
```
stsoe: If clients, aka XRT, are including dtrace.h through explicit submodule path, then how are they going to find `utils.h`? This will not work.  Make sure your exported header files are self contained.

Yes, this is not a new change and xrt_module.cpp already includes
`#include "core/common/aiebu/src/cpp/dtrace/dtrace.h"`
but it ends up including a different utils.h, namely one that is in the compilers include search path.  AIEBU src/cpp/dtrace is not and will not be in the include search path when compiling XRT.

stsoe (changes requested):

src/cpp/dtrace/dtrace.h
AdvaitNaik: My bad, Thanks for catching that, Soren!
Fixed

src/cpp/dtrace/dtrace.h
stsoe: Please only include what you use.

stsoe: How is this macro flipped to dllimport?

stsoe, line 43: why not return the value?

stsoe, line 56: ditto

stsoe: Is this dtrace_buffer in the function?   If so what is the significance of the raw pointer, is it an array?  What's the size?  Is in output?

src/cpp/dtrace/dtrace.h
```diff
@@ -99,21 +97,17 @@ void
 get_dtrace_result_file(dtrace_handle_t dtrace_handle, const std::string& result_file);
```
stsoe, line 83: Misleading function name.  This function doesn't get me anything :-)

src/cpp/dtrace/dtrace.h
```diff
@@ -99,21 +97,17 @@ void
...
+ * and returns the result as JSON string.
  */
-/*
 DTRACE_EXPORT
-void
-update_dtrace_result_buffer(dtrace_handle_t dtrace_handle, const std::string& result_key,
-    nlohmann::ordered_json& result_buffer);
-*/
+std::string
+get_dtrace_result_buffer(dtrace_handle_t dtrace_handle);
```
stsoe, line 95: Good :-)

stsoe (changes requested):

src/cpp/dtrace/dtrace.h
AdvaitNaik: Fixed

AdvaitNaik: DTRACE_EXPORT dates back to the original cert_dtrace SHARED target. The shared library is deprecated and removed; all supported consumers now statically link cert_dtrace_static, so removing the macro

AdvaitNaik: The dtrace APIs currently employ C-style design patterns (output parameters, raw pointers, void returns) due to their original implementation as a dynamically loaded shared library via dlopen(). Now that we have only C++ clients, we should modernize these APIs to follow C++ idioms. 
I will design and implement the updated C++ API in a subsequent PR

src/cpp/dtrace/dtrace.h
AdvaitNaik: Yes, will update the function name with C++ API implementation

PR #314, opened 2026-06-29
aiebu: use $<TARGET_FILE:...> for generator executable commands

mhgerlach30:
CMake's add_custom_command(COMMAND <target>) expands the bare target name as the command string in some generators/configurations, rather than the full path to the built executable.  In a cross-compilation environment the host-built generator binaries are not on PATH, so the command fails with "No such file or directory".

Use the $<TARGET_FILE:tgt> generator expression which always expands to the absolute path of the built executable, making the custom commands work correctly in both native and cross builds.

Compilation problems in cross-compilation environments like BuildRoot.

Use the $<TARGET_FILE:tgt> generator expression in add_custom_commands() making the custom commands work correctly in both native and cross builds.

Change was successfully compiled under a BuildRoot, cross-compile build and a native build under Ubuntu 24.04.

Successfully ran 'ctest -C Release -j4' and 'ctest -C Release -L benchmark' under Ubuntu 24.04.

Request a Windows build test to ensure nothing has been broken.

### DIFF
--- a/lib/src/CMakeLists.txt
+++ b/lib/src/CMakeLists.txt
@@ -42,7 +42,7 @@ file(MAKE_DIRECTORY ${AIEBU_GEN_DIR})
 
 add_custom_command(
   OUTPUT ${AIEBU_PREEMPT_GEN_FILES}
-  COMMAND ${AIEBU_PREEMPTION_EXE}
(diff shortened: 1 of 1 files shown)


stsoe (approved):
LGTM

PR #315, opened 2026-07-01
opcode and patching schema for pl ip to transfer wts

HimanshuChoudhary-Xilinx:
https://amd.atlassian.net/wiki/spaces/AIE/pages/1744115100/Weight+Prefetch+via+PL+IP+and+PLIO+Feature+Specification
Support new opcode and patching schema to support PL IP
Opcode:
1. UC_DMA_MASK_POLL_EXT 
2. APPLY_OFFSET_PL 

Patching schema:
1. pl_ddr_64 

#### TEST
Added PLIP to test all 3.


src/cpp/ops/ops.cpp
```diff
@@ -175,6 +175,27 @@ serialize(std::shared_ptr<assembler_state> state, std::vector<symbol>& symbols,
...
+          sval = std::to_string(val); // buffer_id is the XRT arg index
+          size_t index = state->find_label_entry(m_args[0].substr(1));
+          auto num_entries = state->parse_num_arg(m_args[1]);
+          const std::string ctrltext_patch_sec_name =
+              state->merged_ctrltext_elf()
+                  ? (".ctrltext." + std::to_string(colnum))
+                  : (".ctrltext." + std::to_string(colnum) + "." + std::to_string(pagenum));
+          for (uint32_t numbd = 0; numbd < num_entries; ++numbd)
+          {
+            auto label = state->get_label_at(index);
```
larry9523: what is the expected ASM format when num_entries > 1? Are we expecting multiple labels? Or one label with mulitple wts_params block, or both? Can we please add a test case with num_entries > 1?

larry9523 (commented):
Besides the num_entries comment, I think copilot has an legitimate comment about reuse patching schema 10 from unknown to pl_addr_64. 

While the risk is low, it is still not quite safe if there is any old ELF that has patching schema 10 as a invalid patching when running on new elf parser (xrt). One safe option is to keep 10 as unknown forever and start adding patching schema from 11. Or, we take the low risk to reuse 10, which makes the enum ugly. But to avoid future compatibility issue, we should not use any possible number as unknown. We can choose 0 or -1, for example.

src/cpp/ops/ops.cpp
HimanshuChoudhary-Xilinx: good catch, thought about this found, we can never write a controlcode with num_entries > 1 without tagging them with 2 different labels

